PR #30, opened 2018-01-23
KA10: Add DeCoriolis clock.

larsbrinkhoff:
Here's a pull request made against your latest master.

The clock can be turned on or off.  When off, it reads as zero.  This condition is detected by ITS, which will emit a warning.


larsbrinkhoff:
Resolves #30.

larsbrinkhoff:
Note, if you pause target executing by typing Control-E to enter the simulator console, ITS will be unhappy when execution is resumed.  It will print a BUGPAUSE message.

If I `set pd off` before starting ITS, this message will not appear.

larsbrinkhoff:
See PDP-10/its#643.

PR #39, opened 2018-02-06
KA10: Fix bug in ASHC.

larsbrinkhoff:
Wrong accumulator as source operand.


rcornwell:
Did this change manually.

PR #40, opened 2018-02-06
Automated build and test

larsbrinkhoff:
Adds scripts etc for automated build and test.  The diagnostics are just a stub for now.

You also need to login to Travis CI and enable checking for this repository.


larsbrinkhoff:
It can be a bit of a challenge telling a good run apart from a failed one.  The target machine doesn't have an obvious way to return a status code to the host.

PR #42, opened 2018-02-15
First cut at simulating Type 340 display

philbudne:


rcornwell:
Let me pull over the display directory from the main simh tree and then you can resubmit this pull request.

rcornwell:
I have updated my tree to include the display files from the master SimH tree. I have also updated all the top level SCP files. Please redo your commit, and pull request. All commit messages should start with KA10: 
This is so I can pull them out later.

philbudne:
Do you want the changes to the display directory in a separate commit?

I'll need to update some other simulators, I changed one API (to allow
for 36 console switches instead of just 18).

I figured I would submit a PR with all the display changes to the SIMH.

Phil

rcornwell:
I would recommend only pushing the changes needed for the KA10 here. If you post the other changes to the main simH tree, I can pull them over here. If the changes are only for the KA10, then you can push them here... but I would do them under the KA10 tag.

Remember my goal is to move the commit history over to simH at some point in the future.

philbudne:
My PR has two commits, one labled KA10: the other DISPLAY:
I'm leaving for a weekend away, but if I get a chance, I'll
(cherry-)pick them apart, and submit a new PR.  If not, then
next week.....

rcornwell:
Ok.... What you can do is move the non-KA10 display changes and commit under SCP. But remember they might get wiped out if I update the SCP display directory.  You might want to submit your other changes to simH and I can bring them in that way. You can tag the display changes that are specific to the KA10 340 display as KA10: and that way I can pull them over when I move the repo over to the main SimH tree.

philbudne:
> You can tag the display changes that are specific to the KA10 340 display as KA10: and that way I can pull them over when I move the repo over to the main SimH tree.

That's already the case.  The commit tagged KA10: includes the
ka10_dpy.c file which simulates the Type 348 interface between the
PDP-6/10 I/O bus and the simulated 340 display for SIMH.

I've started work on a commit for the main SimH repo.

rcornwell:
Sorry... misunderstanding. I meant the stuff in display directory that is specifically for the KA as KA10: and the rest of the simH stuff as SCP: The stuff in display/* that is for the KA10 can be tagged with KA10.

philbudne:
Well, everything in the display directory is "generic", not specific
to any host processor (well, except the VT11) or simulator.  Someday
(maybe before another 15 more years down the road), I'd like to make
the Type 340 display available on the PDP-1 and the PDP-4/7/9/15
family.  BUT right now only the "KA" uses the Type 340 simulation.

I put "KA" in quotes, because the Type 348 interface I'm simulating in
ka10_dpy.c is PDP-6 era hardware, but that's exactly what MIT had.
The 3.19 TS Monitor expects this hardware, but by 6.03 it expects the
fancier VB10C which includes DMA, Memory relocation & protection.

I made changes in "core code" so I have to go the long way around for
that, so I might as well for all the changes in the display directory.

PR #58, opened 2018-03-28
KA10: Account for MIT I/O Multiplexor in 340 Display CONI response

philbudne:
Thanks to Lars for figuring it out!

Prevents PDSET "S" command from hanging system, but nothing shows up in display window

### DIFF
--- a/PDP10/ka10_dpy.c
+++ b/PDP10/ka10_dpy.c
@@ -227,8 +227,15 @@ t_stat dpy_devio(uint32 dev, uint64 *data) {
     switch (dev & 3) {
     case CONI:
         *data = (uint64)(uptr->STAT_REG & CONI_MASK);
-        sim_debug(DEBUG_CONI, &dpy_dev, "DPY  %03o CONI PC=%o %06o\n",
-                  dev, PC, (uint32)*data);
+        /*
+         * MIT AI only, See Hardware Memo 1
+         * https://github.com/larsbrinkhoff/its-archives/blob/master/ailab/ITS_Hardware_Memo_1.pdf
+         * Set sign bit if device assigned to this CPU (KA or PDP-6)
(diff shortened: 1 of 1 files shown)


rcornwell:
Can you wait on this pull request? I have major changes to the KA/KI simulators that I have been working on. These changes touch almost every file.

philbudne:
No problem.  Just wanted to make the "fix" (so far) visible.

larsbrinkhoff:
Thanks!  I'll try this in my own copy soon.

larsbrinkhoff:
I think there's a SIMH policy against tab characters.

larsbrinkhoff:
If you want to be picky, you should only set the bit if you previously made a CONO.  Seems like a minor detail, and it probably is, but sometimes ITS is picky about these things.

rcornwell:
Yes there is... I will put up a contribution page next week. If there is tab's in here and this is not a DOS format file, then I will have to reject the pull.

larsbrinkhoff:
No need to reject anything, just update the pull request.

Phil, you can replace the tabs with spaces, amend the commit, and do a forced push to the same branch that was used for the pull request.  I'd do it myself, but I don't have write access to everything on GitHub unfortunately. :-)

rcornwell:
If this is specific to ITS then it probably should be flagged to only occur when ITSPAGING is enabled.

rcornwell:
I would hold off until I push out the changes I have pending. I need to fix ITS.

philbudne:
I amended the commit and force pushed it to philbudne/sims

larsbrinkhoff:
The tabs are gone.  I spotted them when I looked at the diff in the GitHub web interface, and the indentation was slightly off.  (Just like when you view a diff in a text terminal.)

PR #65, opened 2018-05-13
KA10: Fix 340 bug.

larsbrinkhoff:
In the vector function, the flags variable isn't initialized.  The function can sometimes return boolean true even if it shouldn't.

CC @philbudne

### DIFF
--- a/display/type340.c
+++ b/display/type340.c
@@ -453,7 +453,7 @@ vector(int i, int sy, int dy, int sx, int dx)
 {
     struct type340 *u = UNIT(0);
     int x0, y0, x1, y1;
-    int flags;
+    int flags = 0;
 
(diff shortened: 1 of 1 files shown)


philbudne:
Great!  FWIW:  type340.c is in the main SIMH repo
(although the KA10 is the only current use)
I don't know if it makes more sense to submit the fix upstream

larsbrinkhoff:
@rcornwell, would you rather I sent this upstream?

rcornwell:
Yes, send it upstream and I will pull down the latest SCP, it is about time anyway. Right now I am working on sim_card so it will be a couple days.

larsbrinkhoff:
Ok!

PR #67, opened 2018-05-31
KA10: IMX device (A/D converter)

larsbrinkhoff:
I'm making an IMX device stub as per this spec:  
https://github.com/larsbrinkhoff/its-archives/blob/master/ailab/ITS_Hardware_Memo_4.pdf

The stub is just enough to satisfy applicaions there's a something there.  I.e. the CONI assign bit is on, and the DONE is on.

Data passed back to the application will probably be 0 or 2048, and there probably won't be any interrupt capabilities.

### DIFF
--- a/PDP10/ka10_defs.h
+++ b/PDP10/ka10_defs.h
@@ -321,6 +321,7 @@ extern DEVICE   dt_dev;
 extern DEVICE   dk_dev;
 extern DEVICE   pd_dev;
 extern DEVICE   dpy_dev;
+extern DEVICE   imx_dev;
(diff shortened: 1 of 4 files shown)


larsbrinkhoff:
Still no rush with this.  Mostly needed by standalone programs Flight Sim and Dazzle Dart.

Maybe I'll add interrupt capability to make ITS happy, and then PEEK should be able to display the A/D channels and their values.

larsbrinkhoff:
Since Dazzle Dart and Flight Sim are the major applications, I think it would make sense to use joysticks for input to IMX.

larsbrinkhoff:
Oh, I already wrote that above.  I'll delete these useless comments.

PR #66, opened 2018-06-20
KA10: Type 30 window pops up

larsbrinkhoff:
CC @philbudne 

Sometimes the simulator will open a Type 30 window.  For example, this happens reliably when running the standalone [GALAXY](https://github.com/PDP-10/its/issues/932) program.


larsbrinkhoff:
This is now a pull request with a suggested fix.

Tested that this makes GALAXY happy.

PR #74, opened 2018-07-17
KA10: Stub for Rubin 10-11 interface.

larsbrinkhoff:
This is a stub for the PDP-10 side of the Rubin 10-11 interface.  It's a memory mapped device, in the range 3,,000000 to 3,,777777.  (There can be 16K of PDP-6 memory at the start of this range.)

The top page at 3,,776000 is the control page.  It's a pager.  The bottom 256 words describes mappings for the pages in the moby.  Each word has a valid bit, write bit, 3-bit Unibus number, 16-bit Unibus address, and a 10-bit limit.  The limit is the number of PDP-10 words that can be accessed in the page, except the special value 0 which means only one PDP-16 word is accessible.

### DIFF
--- a/PDP10/ka10_cpu.c
+++ b/PDP10/ka10_cpu.c
@@ -216,6 +216,9 @@ int32   qua_tps = 125000;
 #endif
 int32   tmxr_poll = 10000;
 
+/* Physical address range for Rubin 10-11 interface. */
+#define T11RANGE(addr)  ((addr) >= 03000000)
+
 DEVICE *rh_devs[] = {
 #if (NUM_DEVS_RS > 0)
     &rsa_dev,
@@ -356,6 +359,8 @@ MTAB cpu_mod[] = {
 #if ITS
     { UNIT_M_PAGE, UNIT_ITSPAGE, "ITS", "ITS", NULL, NULL, NULL,
               "Paging hardware for ITS"},
+    { UNIT_M_TEN11, UNIT_TEN11, "TEN11", "TEN11", NULL, NULL, NULL,
+              "Rubin 10-11 interface"},
 #endif
 #if BBN
     { UNIT_M_PAGE, UNIT_BBNPAGE, "BBN", "BBN", NULL, NULL, NULL,
@@ -379,6 +384,7 @@ DEBTAB              cpu_debug[] = {
     {"CONI", DEBUG_CONI, "Show coni instructions"},
     {"CONO", DEBUG_CONO, "Show coni instructions"},
     {"DATAIO", DEBUG_DATAIO, "Show datai and datao instructions"},
+    {"TEN11", DEBUG_TEN11, "Show Rubin 10-11 events"},
     {0, 0}
 };
 
@@ -682,6 +688,8 @@ int opflags[] = {
 #endif
 #if ITS
 #define QITS            (cpu_unit[0].flags & UNIT_ITSPAGE)
+#define QTEN11          ((cpu_unit[0].flags & (UNIT_TEN11|UNIT_M_PAGE)) \
+                         == (UNIT_TEN11|UNIT_ITSPAGE))
 #else
 #define QITS            0
 #endif
@@ -1890,6 +1898,15 @@ read:
         sim_interval--;
         if (!page_lookup(AB, flag, &addr, 0, cur_context, fetch))
             return 1;
+#if ITS
+        if (QTEN11 && T11RANGE(addr)) {
+            if (ten11_read (addr)) {
+                nxm_flag = 1;
+                return 1;
+            }
+            return 0;
+        }
+#endif
         if (addr >= (int)MEMSIZE) {
             nxm_flag = 1;
             return 1;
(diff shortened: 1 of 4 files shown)


larsbrinkhoff:
Some outstanding points:

- Unibus is slow compared to PDP-10 memory.  I randomly chose to decrement sim_interval by 100 to reflect this.
- I added some fprintf logging.  This is temporary.

larsbrinkhoff:
The next step is to make some kind of external Unibus interface.  Or rather, eight buses.  And then perform the mappings and forward the requests to the interface.

larsbrinkhoff:
This has been testing by turning on TEN11P in ITS.  ITS will boot and clear all pager words.  It also reads them when PEEK is displaying M or H information (memory/histogram).

Since ITS isn't configured to use any of the three possible perihperal PDP-11s, it will not set up any valid mappings or make accesses to Unibus.

larsbrinkhoff:
I'm not sure which is the best way in SIMH to implement the external interface.  I have a vague sense that TMXR should be used.

larsbrinkhoff:
Hello @rcornwell,

Tell me what needs fixing, at your leasure.

larsbrinkhoff:
I updated the commit to include the address translation logic.

larsbrinkhoff:
Do you want me to check for QITS, in addition to UNIT_TEN11?

rcornwell:
* You should verify that the CPU is in ITS mode before checking access to memory. 
* Remove fprintf and replace them with sim_debug options. Add in a TEN11 option to CPU if you need to.
* UNIT_V_TEN11 should be UNIT_V_MPX + 1

larsbrinkhoff:
I updated as per your instructions.  Let me know if anything's off.

rcornwell:
Shouldn't you check against memory range first?

larsbrinkhoff:
Checking the value of an `AND` versus checking a `>=`.  It's a toss as far as I'm concerned.  Oh, getting the CPU flags may take longer than getting `addr` from a register.  I can swap if you like.

rcornwell:
Did the address have to be within the memory range of the machine or was it above the memory limit? 

In terms of the conditional you should check if you are ITS before you check the range since if ITS is not set, you will never check the address is in the range. The (flags & v) == v should be 3 instructions + branch.

larsbrinkhoff:
The AI pager has a physical address range or 0,,000000 - 3,,777777.  The first three mobies are core, and the last belongs to the 10-11 interface.

The 10-11 "control page" is 3,,776000 - 3,,777777.  The first 256 words control how other pages in the moby are mapped to Unibus addresses.  When an access to the 10-11 moby is made, the page number, i.e. the 776000 bits, are used to look up a word in the control page.

A word from the control page has a Unibus address which is added to the 001777 bits of the physical address.  There's also a limit which says how many words in the page can be accessed.

larsbrinkhoff:
Having said that, I now think I misunderstood your question.

I think the Rubin moby can be anywhere, regardless of how much core the machine has.

rcornwell:
Ok... then prior to checking max memory is fine.

larsbrinkhoff:
I updated the commit to move much of the logic away from the cpu file.  Now it should be much more clear how it works.  It should not be possible to affect TOPS-10 or anything else as long as the TEN11 runtime option isn't enabled.

rcornwell:
This looks much better.

rcornwell:
Lars can you convert this to a device, with one UNIT per unibus connection. Each UNIT can support the attach command to point to the remote 11.

larsbrinkhoff:
Mark directed me to use both TMXR and (host) shared memory for the bus
sharing simulation on the PDP-11 side.  So the KA10 end will have to be
fixed up to do the same.  Since TMXR is involved, there certainly is a
device (the SIMH kind), and maybe some units too.  I haven't read his
changes in detail yet.

larsbrinkhoff:
This pull request is just to intercept the memory accesses and do the address translation.  Can we merge this soon, you think?  I'd like to update ITS and use this in the build.

I have a separate branch for doing the actual communication.

rcornwell:
I would like to see the rest of the code. This code does not add any functionality to the simulator.

larsbrinkhoff:
It allows ITS to run with code enabled for the 10-11 interface.

rcornwell:
Where is the 10-11 interface in this code?

larsbrinkhoff:
As far as the PDP-10 side is concerned, this is the whole interface.  It's all just in the physical memory range 3,,000000-3,,777777.  The top page beginning at 3,,776000 is the pager.  The first 256 words describe how all other pages are mapped to Unibus addresses.

larsbrinkhoff:
I have a follow-up commit in my repo, on the branch lars/unibus.  That adds on top of this, the external interface to talk to a PDP-11 (or whatever).

larsbrinkhoff:
Status update.  I now have the device implemented as a proper SIMH DEVICE, and it uses TMXR for TCP/IP.  Need to clean it up for submission.

larsbrinkhoff:
I'll close this and open another pull request.

PR #80, opened 2018-08-29
KA10: TK10 / NTY / Knight kludge

larsbrinkhoff:
TTY scanner, aka terminal multiplexor.  16 ports.

The device is called both TK10 and Knight kludge.  The code is 600, or NTY symbolically.

### DIFF
--- a/PDP10/ka10_defs.h
+++ b/PDP10/ka10_defs.h
@@ -322,6 +322,7 @@ extern DEVICE   dk_dev;
 extern DEVICE   pd_dev;
 extern DEVICE   dpy_dev;
 extern DEVICE   imx_dev;
+extern DEVICE   tk10_dev;
 extern DEVICE   wcnsls_dev;             /* MIT Spacewar Consoles */
 
 extern t_stat (*dev_tab[128])(uint32 dev, uint64 *data);
@@ -385,6 +386,7 @@ int  df10_write(struct df10 *df);
 #define NUM_DEVS_TU     1
 #define NUM_DEVS_PD     ITS
 #define NUM_DEVS_IMX    ITS
+#define NUM_DEVS_TK10   ITS
 #define NUM_DEVS_DPY    USE_DISPLAY
 #define NUM_DEVS_WCNSLS USE_DISPLAY
 #define NUM_DEVS_IMP    0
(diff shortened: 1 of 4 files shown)


larsbrinkhoff:
This is still a work in progress, but I'm posting a pull request to get eyeballs on it.

It almost works.  The remaining problem is that there are spurious "output ready" interrupts posted.  This makes ITS send garbage i.e. 377 to the terminal.

larsbrinkhoff:
Oh!  The 377s are just padding after CR characters.

Please review this.  In particular, am I using TXMR right?

rcornwell:
Only minor changes. 
  1) Change the debug IRQ to debug Detail then you can remove the tk10_debug structure.
  2) No need for sim_debug calls in attach and detach.
  3) Add in a help function.

larsbrinkhoff:
Question: Is it kosher to name the SIMH device `tk10`?  I see other devices doesn't use numbers in their names.  If, very theoretically, one were to have two units, they would be `tk100` and `tk101`, right?

If this is not a problem (it's unlikely anyone will use more than one), I think `tk10` is good.  `nty` might be another option.

larsbrinkhoff:
Another question.  Is it preferable to have the device diabled from the start?  I noticed I had to do `set tk10 enabled` before attaching it.

larsbrinkhoff:
Call for advice.  When a line isn't connected, should I still call tmxr_putc_ln, or should I check `conn` and drop the character?

larsbrinkhoff:
Using a TK10 line is *very* slow.  Is there a way to change the speed?  I see the global variable `tmxr_poll` is used somehow.  How is that et?

larsbrinkhoff:
Hello Rich,

I updates as per your comments.  I also fixed the speed problem.  You can still set e.g. `speed=300` in the attach command.  Hysterical evidence points to TK10 only operating at a maximum of 300 baud.

This should be ready to go in if `tk10` is an ok device name.

rcornwell:
I don't see a problem with device name of TK10, this might mess up some connect scripts. You could call it TK. And yes this device should be disabled at start. Since it is only for ITS. You probably will want to disable the DC10E in your ITS startup as well.

larsbrinkhoff:
Ok, it's now renamed to TK.

larsbrinkhoff:
Sibling pull request: PDP-10/its#1190.

PR #82, opened 2018-09-18
KA10: Magic switch.

larsbrinkhoff:
Here's the hardware memo: http://www.catb.org/jargon/html/magic-story.html

### DIFF
--- a/PDP10/kx10_cpu.c
+++ b/PDP10/kx10_cpu.c
@@ -230,7 +230,10 @@ uint32  fault_addr;                           /* Fault address */
 uint64  opc;                                  /* Saved PC and Flags */
 uint64  mar;                                  /* Memory address compare */
 uint32  qua_time;                             /* Quantum clock value */
-#endif
+#if MAGIC_SWITCH
+int     MAGIC = 1;                            /* Magic switch. */
+#endif /* MAGIC_SWITCH */
+#endif /* ITS */
 #if KL_ITS
 #define dbr1    FM[(6<<4)|1]
 #define dbr2    FM[(6<<4)|2]
@@ -464,7 +467,10 @@ REG cpu_reg[] = {
     { ORDATAD (OPC, opc, 36, "Saved PC and flags")},
     { ORDATAD (MAR, mar, 18, "Memory address register")},
     { ORDATAD (QUA_TIME, qua_time, 36, "Quantum timer"), REG_RO},
-#endif
+#if MAGIC_SWITCH
+    { ORDATAD (MAGIC, MAGIC, 1, "Magic switch"), REG_FIT},
+#endif /* MAGIC_SWITCH */
+#endif /* ITS */
 #if KL
     { ORDATAD (EXT_AC, ext_ac, 4, "Extended Instruction AC"), REG_HRO},
     { ORDATAD (PREV_CTX, prev_ctx, 5, "Previous context"), REG_HRO},
@@ -4036,6 +4042,13 @@ if ((reason = build_dev_tab ()) != SCPE_OK)            /* build, chk dib_tab */
          reason = STOP_IBKPT;
          break;
     }
+
+#if MAGIC_SWITCH
+    if (!MAGIC) {
+         reason = STOP_MAGIC;
+         break;
+    }
+#endif /* MAGIC_SWITCH */
 
     check_apr_irq();
     /* Normal instruction */
--- a/PDP10/kx10_defs.h
+++ b/PDP10/kx10_defs.h
@@ -86,6 +86,10 @@
 #define PDP6_DEV PDP6|WAITS
 #endif
 
+#ifndef MAGIC_SWITCH   /* Infamous MIT magic switch. */
+#define MAGIC_SWITCH 0
+#endif
+
 
 /* MPX interrupt multiplexer for ITS systems */
 #define MPX_DEV ITS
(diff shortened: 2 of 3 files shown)


larsbrinkhoff:
Here's a photo of this important device.

magicswitch1

larsbrinkhoff:
I updated this pull request.  The switch is now a register instead of a modifier.  It now exits the simulator loop.

larsbrinkhoff:
@rcornwell, is the second version better?  I'll squash the two commits together.

rcornwell:
Lars can you please rebase this against current build.

larsbrinkhoff:
Rebased now.

larsbrinkhoff:
Rebased again.  Anything needed done here?

rcornwell:
Can we close this, I really don't see that this is a good addition to the simulator since it adds no functionality to the simulator and only reflects and idiosyncrasy of one machine?

larsbrinkhoff:
To answer your question, yes it was only on one machine.

larsbrinkhoff:
It seems the magic switch is plannel to be part of the PiDP-10.  So I will hang on to this code.

larsbrinkhoff:
I rebased to kl10, so there are a lot of commits in this pull request that shouldn't be there.  That will be cleaned up once the kl10 branch is merged to master.

@PiDP, this is my current implementation of the magic switch.  You type `dep magic 0` to reduce the amount of magic, which breaks out to the SIMH prompt.

rcornwell:
I am still reluctant to add this to the simulator. Since it's only purpose is to halt the simulator. If Oscar wishes to install this switch in back of his simulator, that is possible. You need to have magic == 0 for simulator to run normal, and magic != 0 to report no magic.

@larsbrinkhoff  if you want this included, you will need to put in a #define MAGIC_SWITCH 0 into the defs file, and have everything selected with #if MAGIC_SWITCH. This way the default is to not compile this, but if the user desires this functionality can be added. Also this only really makes sense in the case of real hardware. Note the halt switch preforms the exact same function! But please wait until the KL10 branch is merged into the main tree and I have done cleanup to code.

larsbrinkhoff:
Yes, you are right that having a halt switch and a magic switch doing the same thing is silly.  I don't know what Oscar has in mind.

Anyway, I'm happy to make those changes.

PiDP:
Richard,

On Fri, 14 Feb 2020 at 16:31, Richard Cornwell <notifications@github.com>
wrote:

> I am still reluctant to add this to the simulator. Since it's only purpose
> is to halt the simulator.
>

Indeed, for me there is no need for software changes. Making some sort of
Magic Switch will be electronically interesting (as it turns out, it is not
simple to devise a circuit that will detect a switch change just on effects
in the single wire). The 'fun' therefore is in the electronic side of this
gadget feature. I can easily make the Raspberry Pi translate the electric
signal into a simh command, no need to accommodate it in the software
explicitly.

There's a few people currently thinking of the best way to make a switch on
a single wire do something. It's a nice electronics puzzle! And as the
PiDP-10 circuit board has acres of unused space, why not add a Magic
circuit...

The common explanation - that somehow, the case of the switch conducted to
Frame Ground - is not really a plausible one. The switch in question seems
to be a pretty high-end type, and the idea that it developed a fault where
the metal casing somehow was connected to one of its three signal pins is
very unlikely. At the moment, I'm playing with the idea of a tiny switch
panel-cum-circuit board where traces on the circuit board (not the switch
innards, they are too small) cause either a capacitive or signal-bouncing
effect, as a metal case would. Much like you can damage a HF transmitter by
transmitting without an antenna, it is possible to have a feedback
influence from that on a high-frequency signal.


Kind regards,

Oscar.

PiDP:
Lars,

On Fri, 14 Feb 2020 at 17:46, Lars Brinkhoff <notifications@github.com>
wrote:

> Yes, you are right that having a halt switch and a magic switch doing the
> same thing is silly. I don't know what Oscar has in mind.
>

It's not so much the functionality (although a crash switch seems useful
for many computer applications, such as nuclear launch facilities for
instance), but the electronics challenge of making a switch on a single
wire do something. Call it an electronics folly!

Kind regards,

Oscar.

rcornwell:
Looking at the picture of the back of the switch, you can see it was not actually a one wire switch. There is one wire which I believe goes to some pin on the KA10 backplane. (note) would be interesting to know where it actually connected. And a second wire wrapped around the metal standoff. This would act as an antenna feeding noise into the circuit. My guess is that the metal post was connected to the frame, which brought it to ground. In one position there was no connection to ground. In the other position there was a small inductive/capacitive path to ground which loaded the circuit down. It is also possible that there was increased AC coupling when someone touched the switch causing noise to go on the bus, thus crashing the 10.

My feeling still stands, I do not believe that this belongs on either a simulator nor a simulator in a case.

larsbrinkhoff:
I understand your reservations, in so far it doesn't add anything useful.  I'm not disputing that.

On the other hand, what's the harm?  It's part of hacker folklore and may add a little something to the ITS experience.

I'm not twisting your arm.  If your suggested edits make it OK for you, I'll do them.  If you're dead set against it, I'll close the pull request.  But it may pop up in the ITS repository.

rcornwell:
Rebases this and add in to kx10_defs.h
`
#ifndef MAGIC_SWITCH
#define MAGIC_SWITCH 0
#endif

#if MAGIC_SWITCH & !KA & !ITS
#error  "Magic switch only valid on KA10 with ITS mods"
#endif
`

Then define enclose all MAGIC switch stuff in #if MAGIC_SWITCH. This means that if the user wishes they can add this definition to their compile command.

larsbrinkhoff:
No problem.  Will do when DPK, TK, and IMX are updated.

larsbrinkhoff:
Rebased and updated.

rcornwell:
The #ifndef MAGIC_SWITCH needs to be near the top of the file with all the other definitions. Where you have it now it will result in never defining the stop code for magic.

larsbrinkhoff:
Updated and tested.

rcornwell:
That looks better I will pull this in tonight.

Issue #64, opened 2018-05-11
KA10: .DSTART crash


larsbrinkhoff:
The test program below will reliably crash ITS in one of two ways.  If run directly, ITS will hang.  If run under DDT, here's a BUGHLT.

The display list passed to .DSTART is actually wrong.  But I don't think that should crash ITS.

```
a==1
p==17

go:     .dstart [-dlen,,dlist-1]
         .lose
        movei a,5*30.
        .sleep a,
        .logout 1,

dlist: 373737,,3000    ;Escape, Stop
dlen=.-dlist

end go

**Comment by larsbrinkhoff** [comment]
CC @philbudne on this.  (There's no new information in this issue since I last posted about this.)

**Comment by larsbrinkhoff** [comment]
Hmmm, now I can't get the BUGHLT from DDT.  But I *think* it looked like this:

```
UUO WITH PI IN PROGRESS  
PI LEVEL 6 BUGHALT.  FIND A WIZARD OR CONSIDER TAKING A CRASH DUMP.

larsbrinkhoff:
Apparently there's something about the argument to .DSTART.

- If the BLKO pointer has a small negative number in the left half, ITS is prone to crash.
- If it's something like `DLIST-2000`, it's ok (this is copied from PDSET).
- I tested -2000, and got a DPY interrupt (but no crash).

larsbrinkhoff:
I have compared a "small negative number" run against "dlist-2000".  Everything seems to go well until the 340 interrupt service routine calls LPTRT1 to exit.  When the JRST 12, instruction is executed, the 340 is supposed to interrupt but it doesn't.  Instead ITS goes into the idle 17/ AOJA 17 loop from which it never returns.

larsbrinkhoff:
CLKTIC isn't called.

I see console TTY interrupts are executed.

larsbrinkhoff:
Here's a log with CPU DEBUG=IRQ.
[irq.log](https://github.com/rcornwell/sims/files/2136692/irq.log)

larsbrinkhoff:
From this passage, I think dpy_svc raises an interrupt (that line is my own addition).

Then the interrupt is serviced several times.  It doesn't look like DPY is clearing the interrupt.  Maybe that's wrong, or maybe it's intentional.

Then dpy_update_status raises an interrupt.  At this point, there already is an interrupt active, so "set irq" says PIH is 002.

Then the interrupt is cleared.  There is now a TTY interrupt pending (010).

[...]

PR #85, opened 2018-10-05
KA10: Type 340 device is flickering

larsbrinkhoff:
Some 340 applications running in timesharing displays a badly flickering view.  These include FED, SPCWAR, and MacHack VI (the program eloquently named C).


larsbrinkhoff:
I'm posting a pull request for review and discussion.

I remember from #64 I saw some problems with the device interrupt handling.  My first commit rearranges the interrupt handling to make sure an interrupt is flagged whenever the DONE or SPECIAL status bits are on, and is cleared when they are both off.  Also, the delay for the DONE interrupt is changed so that the bit is not set until the delay has expired.

This is a marked improvement for FED.  The second commit tunes down the delay which makes the display quite stable.

However, SPCWAR and C are still flickering badly.

larsbrinkhoff:
I tested PEEK just now.  Before my change, it was prefectly fine.  But after, it's bad.

PDP10/ka10_dpy.c
```diff
@@ -180,10 +180,24 @@ const char *dpy_description (DEVICE *dptr)
...
+    uptr->DELAYED_DONE = 1;
     uptr->INT_COUNTDOWN = INT_COUNT;
 }
 
+/* update interrupt request */
+static void check_interrupt (UNIT *uptr)
+{
+    if (uptr->STAT_REG & CONI_INT_SPEC) {
+        uint32 sc = uptr->STAT_REG & CONX_SC;
+        set_interrupt(DPY_DEVNUM, sc >> CONX_SC_SHIFT);
```
philbudne: Is it OK to call set_interrupt with a zero value? Same thing at line 195.  Would be nice if the comment at 187 explains who calls this.  The name "check_interrupt" sounds a little like this is about interrupt service, when it's about sending interrupts.

larsbrinkhoff: Good points.  Yes, that should be ok.  The set_interrupt function checks for zero itself.

larsbrinkhoff:
I'm closing this to make it clear the branch is not in a good state.

larsbrinkhoff:
This shows MacHack flickering quite a bit, at 5:45.  
https://www.csail.mit.edu/node/6666

larsbrinkhoff:
A random thought.  If DISPLAY_CYCLE_US is 50, that means the svc function is called every 50 microseconds, or thereabouts.  On top of that, ka10_dpy.c counts down INT_COUNT times before triggering an interrupt.  The default is (500/DISPLAY_CYCLE_US), so every 0.5 milliseconds.  This goes for data interrupts, not special interrupts. 

I wonder if this rate is right?  Consider point mode.  One of those instructions are documented to take 3 𝜇s for data transfer (an 18-bit instruction?), 35 𝜇s for positioning, and 0.5 𝜇s for intensification.  The total is 38.5 𝜇s, or 77 𝜇s per 36-bit word.

Or consider vector or incremental mode.  The time per point is 3 𝜇s + 1.5 𝜇s.  I think that goes for each point in a vector, so a, say, 200-point vector would be 1.8 ms.  Or a whopping 3.6 ms if there are two such vectors in a PDP-10 word.

I don't see how long it takes to draw a character (but I didn't look too hard).  I think a reasonable guess is to treat is at 5x7 incremental points, so 3 𝜇s + 35x1.5 𝜇s = 55.5 𝜇s.

larsbrinkhoff:
According to a PDP-7 manual, 35 𝜇s.

PR #87, opened 2018-12-03
KA10: Rubin 10-11 interface.

larsbrinkhoff:
The Rubin 10-11 interface is shared memory between a PDP-10 and eight Unibuses.


rcornwell:
Only change I would make to this is to pass a pointer to MB as the second argument to ten11_read and ten11_write. Not sure exposing MB to all modules is a good idea.

PR #88, opened 2018-12-11
KA10: Stanford keyboard

larsbrinkhoff:
The MIT AI Lab PDP-6/10 had a Stanford keyboard attached as an I/O bus device.  ITS has support for this.

There was a Datapoint terminal emulator that displayed on the Type 340, and used the Stanford keyboard for input.

With this in mind, it would make sense to take keyboard input from the simulator Type 340 window.

Keyboard layout: http://www.keyboard-layout-editor.com/#/gists/6000ae5d36813107c38e6396a5acd518

SAIL keyboard

### DIFF
--- a/PDP10/ka10_dpy.c
+++ b/PDP10/ka10_dpy.c
@@ -244,7 +244,7 @@ t_stat dpy_devio(uint32 dev, uint64 *data) {
         uptr->STAT_REG &= ~CONO_MASK;
         uptr->STAT_REG |= *data & CONO_MASK;
         if (*data & CONO_INIT)
-            dpy_update_status( uptr, ty340_reset(), 1);
+            dpy_update_status( uptr, ty340_reset(&dpy_dev), 1);
         sim_debug(DEBUG_CONO, &dpy_dev, "DPY %03o CONO %06o PC=%06o %06o\n",
                   dev, (uint32)*data, PC, uptr->STAT_REG & ~STAT_VALID);
         break;
@@ -298,7 +298,7 @@ t_stat dpy_reset (DEVICE *dptr)
 {
     if (!(dptr->flags & DEV_DIS)) {
         display_reset();
-        ty340_reset();
+        ty340_reset(dptr);
     }
     sim_cancel (&dpy_unit[0]);             /* deactivate unit */
     return SCPE_OK;
--- a/PDP10/ka10_defs.h
+++ b/PDP10/ka10_defs.h
@@ -323,6 +323,7 @@ extern DEVICE   pd_dev;
 extern DEVICE   dpy_dev;
 extern DEVICE   imx_dev;
 extern DEVICE   imp_dev;
+extern DEVICE   stk_dev;
 extern DEVICE   tk10_dev;
 extern DEVICE   mty_dev;
 extern DEVICE   wcnsls_dev;             /* MIT Spacewar Consoles */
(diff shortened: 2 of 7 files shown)


larsbrinkhoff:
This isn't tested yet, but essentially everything is there.

larsbrinkhoff:
I added a commit from mainline SIMH.  It can go away when the master branch has been updated with the latest SCP changes.

larsbrinkhoff:
I added code to use @markpizz's new `vid_display_kb_event_process` API.

I'll let this pull request rest until @rcornwell has updated this repository with the latest SIMH changes.

larsbrinkhoff:
Done.

larsbrinkhoff:
Rebased; not yet retested.

larsbrinkhoff:
Good to go as far as I can see.

This is a physical key mapping, so e.g. typing 0 gets you + since that's what the Stanford keyboard had to the right of the 9.

PDP10/ka10_dpy.c
```diff
@@ -243,8 +243,10 @@ t_stat dpy_devio(uint32 dev, uint64 *data) {
         clr_interrupt(dev);
         uptr->STAT_REG &= ~CONO_MASK;
         uptr->STAT_REG |= *data & CONO_MASK;
-        if (*data & CONO_INIT)
+        if (*data & CONO_INIT) {
             dpy_update_status( uptr, ty340_reset(&dpy_dev), 1);
+            ty340_set_grid(6, 9);
```
philbudne: I thought you had renamed this function ty342_set_grid?
Since (for now) the PDP-6/10, running MIT s/w is the only client, any reason not to make this the default?
OR if you keep the call in ka10_dpy, make a function that calls ty340_reset, sets the spacing and returns the 340 status??

larsbrinkhoff: Yes, I renamed it in the SIMH pull request.  Thanks for the reminder, I should rename it here too.

The reason would be that the only documentation we have, for PDP-7, insists it's 7x12.  Maybe it was configured differently for the PDP-6, or MIT made a local modification.  I'm happy to make 6x9 the default; I think it's your call.

larsbrinkhoff:
Not good to go because it now depends on upstream SCP changes which haven't been merged yet.  Still, it would be good if Rich could give an indication if the pull request needs any modifications.

philbudne:
I'd make 6x9 the default.  The PDP-7 manual also has hand written corrections,
so it was clearly preliminary.  If someone discovers software for smaller
machines that needs something else, the set grid call is there for them!

p

rcornwell:
SCP is up to the current level. I will try and pull changes in over the weekend. Work has been busy this week.

larsbrinkhoff:
Thanks.

I have a pull request pending for SIMH, which this pull request depends on.

larsbrinkhoff:
> Since (for now) the PDP-6/10, running MIT s/w is the only client, any reason not to make this the default?

@philbudne, FYI, I think some 340 software from UPenn has survived too.  But I checked, and they didn't have the 342.

larsbrinkhoff:
Mark meged my "sim_ws" pull request upstream.

rcornwell:
In the future before submitting patches, please verify that both KA10 and KI10 build. Also verify that they build without TYPE340=1!

larsbrinkhoff:
Thanks, I'll keep that in mind.

rcornwell:
Merged.

PR #93, opened 2019-01-15
KA10: Auxiliary PDP-6 processor

larsbrinkhoff:
Simulate MIT I/O device 20 and shared memory with a PDP-6.

The device is used to send I/O bus interrupts between the processors.  Shared memory works much like the 10-11 interface.  The KA10 can read and write PDP-6 core, but not vice versa.

Work in progress.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,14 @@
+os:
+  - linux
+#  - osx
+language: c
+dist: trusty
+sudo: required
+env:
+  - CPU=ka10 TYPE340=0
+  - CPU=ka10 TYPE340=1
+  - CPU=ki10 TYPE340=0
+install: sh -ex dependencies.sh install_linux
+script: make $CPU TYPE340=$TYPE340
+notifications:
+  email: lars@nocrew.org
--- a/PDP10/ka10_auxcpu.c
+++ b/PDP10/ka10_auxcpu.c
@@ -1,0 +1,391 @@
+/* ka10_auxcpu.c: Auxiliary processor.
+
+   Copyright (c) 2018, Lars Brinkhoff
+
+   Permission is hereby granted, free of charge, to any person obtaining a
+   copy of this software and associated documentation files (the "Software"),
+   to deal in the Software without restriction, including without limitation
+   the rights to use, copy, modify, merge, publish, distribute, sublicense,
+   and/or sell copies of the Software, and to permit persons to whom the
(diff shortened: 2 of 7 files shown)


rcornwell:
Sorry, but not going to pull this one in. It will break later PDP6 support. And be very confusing for the user of the simulator.

larsbrinkhoff:
It's your choice.  Granted it could be confusing, but it wouldn't have to break anything.

It's regrettable, but the ITS repo will have to run off a fork of your KA10 simulator.

rcornwell:
If you want this feature added you need to address the confusing which will occur when PDP6 support is added to KA10. There will be a "set cpu pdp6" command added in the near future. Having a PDP6 device will be confusing to the novice user.

larsbrinkhoff:
Sorry, I thought you wanted nothing of this.  I can certainly rename it.  It's not exactly specific to PDP-6, it could be any PDP-10 with the right memory bus.

Some random ideas for other names:

- Shared memory, SH, SM, SHMEM.
- Slave, SLV.
- Auxiliary, AUX.
- Core... maybe not so good.

larsbrinkhoff:
Also note:

> Work in progress.

I'm rewriting this to use TMXR like ten11.  It's not finished.

rcornwell:
I like AUXCPU or similar. Note this does not have to specifically be related to ITS. The TYPE340 should also work with KI. Let me know when you have it cleaned up. I am still working on getting IMP cleaned up.

rcornwell:
Yes, I will be moving this into the page function I think. Along with converting memory access to indirect functions.

larsbrinkhoff:
I'll await the carnage with trepidation. :-)

larsbrinkhoff:
@aap found a bug: The auxcpu line wasn't properly closed when the connection went away.  I fixed that, and also did the same thing for ten11.

rcornwell:
I pushed out the changes for Memory access functions. If you update Auxcpu to the new format I will merge it later this week. You might also be able to eliminate some of the testing in ITS routines to improve performance. Mem_read_its and Mem_write_its will only ever be called if the CPU is defined as an ITS CPU. I would reverse the tests for TEN11 and Address range check, so you check if it is in range before see if 11 is installed.

larsbrinkhoff:
Updated.  Not yet tested.  Also, I'll remove the Travis CI commit.

larsbrinkhoff:
Did some light testing.

PR #95, opened 2019-03-06
KA10: Reverse TEN11 checks.

larsbrinkhoff:
As per request in #93.


rcornwell:
Does this change improve performance any or is it not noticeable?

larsbrinkhoff:
I haven't checked.  Just swapping the checks I'd guess doesn't matter much, especially if you're talking to something across the network.

rcornwell:
It does not matter for the case where you are talking over the network, however 99% of your memory accesses will not be over the network, but to local memory. This saves a mask and conditional branch for each memory access. This way you only check if this really is a PDP11/PDP6 access after you decide you need to do it.

larsbrinkhoff:
Makes sense, I'm all for it!

rcornwell:
I manually added this patch in so I am closing this push request.

PR #98, opened 2019-04-05
I7000: Fix MT density management

markpizz:
This device supports two densities.  Both can be manually set with a
"SET MTA0 LOW|HIGH" or under program control.

This change seems to allow sim_tape to be aware of the density
that the simulated system is operating under.


rcornwell:
Looks good, I will pull this in in a couple days. I have bunch of changes for PDP6 support pending.

Issue #110, opened 2019-06-12
KA10: multiplication


larsbrinkhoff:
The ITS file MAINT; PART F.OLD has these checks:

```
5025/  MOVSI 12,400000
5026/  HRROI 11,777777
5027/  MUL 12,11
5030/  JOV 5032

5034/  MOVSI 12,400000
5035/  HRROI 11,777776
5036/  MUL 12,11
5037/  JOV 5041
```

There isn't solid evidence, but my assumption is that these checks are for a KA10.  The simulator fails these checks.  The simulator passes PART G.OLD and PART K.OLD.

larsbrinkhoff:
The 82 ProcRefMan says:

> CAUTION
> 
> In the KA10, an AC operand of -2<sup>35</sup> is treated as though it were +2<sup>35</sup>, producing the incorrect sign in the product.

larsbrinkhoff:
Testing on the LCM machines:

| CPU | Que pasa?
| --- | ---
| KA10 | Overflow
| KI10 |
| KL10 | **No overflow**
| KS10 | Overflow

larsbrinkhoff:
Experimental results on the KA10:

| AC input | Memory intput | AC output | AC+1 output | Flag
| -- | -- | -- | -- | -- 
| 400000000000 | -2 | -2 | 400000000000 | Overflow
| 400000000000 | -1 | -1 | 400000000000 | Overflow
| 400000000000 | 0 | 0 | 0 | -
| 400000000000 | 1 | 1 | 0 | -
| 400000000000 | 2 | 2 | 0 | -

larsbrinkhoff:
I think this passes DEC diagnostics.

```
diff --git a/PDP10/ka10_cpu.c b/PDP10/ka10_cpu.c
index e3b8fcf..a782438 100644
--- a/PDP10/ka10_cpu.c
+++ b/PDP10/ka10_cpu.c
@@ -4128,6 +4128,10 @@ left:
               }
               AR &= FMASK;
               MQ = (MQ & ~SMASK) | (AR & SMASK);
+#if KA
+              if (BR == SMASK && (AR & SMASK))  /* Handle special case */
+                 FLAGS |= OVR;
+#endif
               break;

     case 0230:       /* IDIV */

larsbrinkhoff:
It also passes MAINT; PART F.OLD.

rcornwell:
This patch is in current commit.

larsbrinkhoff:
[...]

Issue #109, opened 2019-06-05
KA10: IDIV -2^35 / -1


larsbrinkhoff:
I believe there is a problem with IDIV -2<sup>35</sup>/-1.

First, as per documentation:

> If [...] AC contains -2<sup>35</sup> and the operand specified by M is -1 (except in the KS10), set Trap 1, Overflow, and No Divide, and go directly to the next instructions without affecting the original AC or memory operand in any way.

https://archive.org/details/bitsavers_decpdp1019_38003019/page/n63

Second, running on the LCM KA10 (TOPS-10):

```
.DEBUG TEST
LINK:   Loading
[LNKDEB DDT Execution]

START/  MOVSI 1,400000  
DDTEND+1/       SETO 2,0        
DDTEND+2/       IDIVM 1,2       
DDTEND+3/       JRST 4,0        .$B     

$G
$1B>>DDTEND+3   
1/      SETZ 0  
2/      -1      
```

But on the simulator (ITS):

```
100/   MOVSI 1,400000   
101/   SETO 2,   
102/   IDIVM 1,2   
103/   0   .$b   

100$g
$1B; 103>>0   0/   0   0/   0      
1/   SETZ   
2/   SETZ
```

larsbrinkhoff:
I searched diagnostics to see if I could find a test of this particular corner case, but didn't find anything obvious.  I looked at DEC diagnostics files called DBKDAM ,DFKDAM, DSKAL, DBKALM, and one in ITS called MAINT; PART F.

rcornwell:
KA diags are DAKxx DBKxx are for KI10. I am not sure that the KA had this issue. KA could not trap so this would be for KI or KL.

larsbrinkhoff:
No Divide is also triggered by -2<sup>35</sup>/+1, according to ProcRefMan.  This goes for both KA and KI.

rcornwell:
What does the diags say about this specific case?

larsbrinkhoff:
I don't know my way around the diagnostics, but I'm keen to find out.  I'll let you know if I find anything.

rcornwell:
Once I get FP working correct I will try and add this trap and see if diags complain.

larsbrinkhoff:
Cool.

I believe DAKAL is the KA10 division diagnostics.  As far as I can see there's no check for a dividend -2<sup>35</sup> and divisor ±1.  I have looked around all IDIV and DVND in this file:

[...]

PR #111, opened 2019-06-12
KA10: Fixes for IDIV and MUL.

larsbrinkhoff:
Some quirks on the KA10 and PDP-6 processor.  See #109 and #110.


rcornwell:
These fixes are in latest commit.

PR #124, opened 2019-07-07
I7090: Fix test script so that it runs to completion without errors.

markpizz:
Not all platforms allow an open file to be deleted.

Now the explicit errors are removed and the script will run identically on all platforms.
What is missing is a way to specifically test for expected results and to confirm that they have actually happened.  Comparison about expected halt PC or other expected register results or output would be appropriate and simulator exit on failure.


rcornwell:
I manually merged in this push.

Issue #12, opened 2018-01-14
Add SA10 device support for Tymcom-x


rcornwell:

b4:
This may also benefit CompuServe

rcornwell:
I need documentation on the device to be able to implement it. Also do we have images or sources for CompuServe system?

b4:
Images, yes.

Source? No

rcornwell:
What machine is the image for?

b4:
KL/SC-40

rcornwell:
I am not sure if I will do support for a KL10. I have no plans to do a SC-40. Tymcom-x supports a KI.

larsbrinkhoff:
Maybe parts of the SA10 support could be moved over to KLH10.

b4:
Some documentation: https://github.com/PDP-10/tenex/blob/master/doc/sa10f.doc
https://github.com/PDP-10/tenex/blob/master/133-tenex/sa10d.mac
https://github.com/PDP-10/tenex/blob/master/cusps/sa10f.mac
https://github.com/PDP-10/tenex/blob/master/cusps/saio.mac
https://github.com/PDP-10/tenex/blob/master/cusps/sacopy.mac
(related documentation in doc/)

https://github.com/PDP-10/tenex/blob/master/doc/pilevels.txt#L8

larsbrinkhoff:
CMH has some SA-10 documentation: https://www.computerhistory.org/collections/catalog/102748567

CC @kossow

larsbrinkhoff:
> I am not sure if I will do support for a KL10.

Update: KL10 now supported, both models A and B.

SA10 docs still at large.

larsbrinkhoff:
It's up now.  Thanks @kossow!
http://www.bitsavers.org/pdf/systemsConcepts/SA10/SA10_Users_Manual_197503.pdf

larsbrinkhoff:
Diagnostics: SA10D.MAC and SA10F.DOC in https://github.com/PDP-10/tenex

larsbrinkhoff:
More diagnostics? https://github.com/TYMCOM-X/169273.tape/blob/abc68e373db6be0104efe986d23624462ef691b9/diag/sa10xx.mac

PR #116, opened 2019-07-11
KA10: Systems Concepts DC10 disk controller

larsbrinkhoff:
One-off device for the AI KA10.  (Not to be confused with the DC10 data line scanner.)  Useful for running binary-only version of ITS version 724 from 1972.

I'm guessing there may be some similarities to the SA10 (see issue #12).

### DIFF
--- a/PDP10/ka10_ai.c
+++ b/PDP10/ka10_ai.c
@@ -1,0 +1,1035 @@
+/* ka10_ai.c: Systems Concepts DC-10 disk control
+
+   Copyright (c) 2019, Lars Brinkhoff
+
+   Permission is hereby granted, free of charge, to any person obtaining a
+   copy of this software and associated documentation files (the "Software"),
+   to deal in the Software without restriction, including without limitation
+   the rights to use, copy, modify, merge, publish, distribute, sublicense,
+   and/or sell copies of the Software, and to permit persons to whom the
+   Software is furnished to do so, subject to the following conditions:
+
+   The above copyright notice and this permission notice shall be included in
+   all copies or substantial portions of the Software.
+
+   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+   IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+   LARS BRINKHOFF BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+   IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+   CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+   This disk controller was probably only ever used with the MIT AI
+   Lab PDP-10.  Since the device name DC is alreay claimed, we call
+   this AI.
+*/
+
+#include "kx10_defs.h"
+
+#ifndef NUM_DEVS_AI
+#define NUM_DEVS_AI 0
+#endif
+
+
+#if (NUM_DEVS_AI > 0)
+
+/* Disk pack geometry.  The track format is software defined.  ITS and
+   SALV makes it hold two sectors with 1024 words of regular data and
+   4 extra words. */
+#define SECTOR_SIZE        1024
+#define SECTORS            2
+#define SURFACES           20
+#define MEMOREX_CYLINDERS  203
+#define CALCOMP_CYLINDERS  (2 * MEMOREX_CYLINDERS)
+#define CYLINDER_SIZE      (SECTOR_SIZE * SECTORS * SURFACES)
+#define MEMOREX_SIZE       (CYLINDER_SIZE * MEMOREX_CYLINDERS)
+#define CALCOMP_SIZE       (CYLINDER_SIZE * CALCOMP_CYLINDERS)
+
+/* The real sector size, including 2 header words, 4 extra data words,
+   and 2 checksum words. */
+#define SECTOR_REAL_SIZE   (SECTOR_SIZE + 8)
+/* A track actually has some more free space.  Cylinder 0, surface 0
+   has a readin block there. */
+#define TRACK_REAL_SIZE    ((SECTORS + 1) * SECTOR_REAL_SIZE)
+#define CYLINDER_REAL_SIZE (SURFACES * TRACK_REAL_SIZE)
+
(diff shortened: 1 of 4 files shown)


larsbrinkhoff:
I'm cleaning it up.  I'm not considering it finished, but I'm posting it here anyway.

If you have time, please comment.

rcornwell:
A couple comments:
1) Change the name to ka10_aidc.c or ka10_dc10.c, to avoid potential confusion with kx10_dc.c, internal names should match device names. This makes debugging easier.
2) Be careful with assign int = uint64, this is not portable to all platforms.
3) Do NOT used v = (data & mask) | (-1 < n) to sign extend a number, this is not portable. The preferred method is to count up until mask+1 gets set.
4) According to Mark TYPE340 is built by default if you have SDL installed. But top level changes should not be in same push request
5) Change NUM_DEVS_AI to NUM_DEVS_AIDC to be consistent.
6) #define the geometry of the disk at the top of the file. This will make code easier to maintain. Or place geometry into structure table.

When it is working let me know I will pull it in.

larsbrinkhoff:
Thanks.

About 3, I'm not sign extending.  I indend to make a negative number.  I could use ~0 if you prefer.

What do you mean with "count up until mask+1 gets set"?

larsbrinkhoff:
I'm still struggling with the name.  AIDC doesn't exactly roll off the tongue.  DC10 would be nice, but then there are units DC100, DC101, etc.  I was toying with the idea to just name it AI.  Yes, I will make it consistent, but consistent with what?

I'd prefer SCDC over AIDC.  Or just AI.

rcornwell:
I am not as concerned as to what you call it, just make the name consistent in the file and the device.

For number 3 I know what you are trying to do. Typical 10 stuff is a negative word count. You want to make it negative so you can then do a -(a) to get a positive word count. If you have a negative work count of n bits... i.e. a mask of n 1 bits. For each word you transfer you increment the word count. When the word count becomes 0 the number will equal mask + 1 (1 and n 0's).

Also when you do -((n & mask) | (-1 < m)) this will be incorrect for the 0 case! The zero case means to transfer mask number of words. 

Also I just noticed this, do not transfer &M[data & ADDR], ... , n, file) this can result in a segment fault or worse overwriting core! Transfer a sector or a track to a buffer and then write it. This will allow for proper address wrap.

aidc_svc should process one command, which processes one word and then reschedules itself. The way you are doing it could cause the simulator to get stuck in a loop.

Take a look at kx10_rc.c or pdp6_dsk.c to see how to do word based transfer.

larsbrinkhoff:
Updated to count up to mask+1, or use "mask+1 - count" where appropriate.

larsbrinkhoff:
Changed device name to AI throughout.

larsbrinkhoff:
SALV does something special to cylinder 0, surface 0.  There are two ordinary sectors 0/0/0 and 0/0/1.  However, on this track only, there is a third sector which is only 16 words and it's also tagged 0/0/0.  This is called the *readin* block.

If I understand DSKDMP correctly, it will write a channel program to this sector.  The channel program writes and reads two sectors.  It's unclear what's supposed to happen next.

larsbrinkhoff:
Added NXM checks.

larsbrinkhoff:
Typical test procedure I have used:

1. Start SALV from MAGDMP.
2. Run MARK to format a disk pack.
3. Run TRAN to write some files.
4. Run LISTF to list files.
5. Run RDHEAD to decode sector headers.
6. Reboot into DSKDMP.
7. Load and run ITS.
8. Log in, list files, shut down.

larsbrinkhoff:
> aidc_svc should process one command, which processes one word and then reschedules itself. The way you are doing it could cause the simulator to get stuck in a loop.

The svc function calls channel_command to process one instruction in the channel program, so it can not get stuck in a loop.

Sorry, I don't want to rewrite this to transfer single words.

larsbrinkhoff:
I'd like to put this through it paces with a full ITS repository build.  I will report back here when done.

rcornwell:
I think you should set up DC10 to transfer 1 word every 10 or so sim_interval cycles. This will clear up many of the potential errors and stability issues you might introduce.

larsbrinkhoff:
Got through a full ITS build with DC10.  Rebased and squashed the commits.

Please merge this.

larsbrinkhoff:
http://www.bitsavers.org/pdf/ibm/dasd/2314/A26-3599-4_2314_Sep69.pdf

larsbrinkhoff:
I understand you have some reservations about the implementation.  I think this is an appropriate level of simulation for this device.  You don't, and that's OK.  Sometimes it's better to agree to disagree.  If you don't want this code, I will respectfully withdraw the pull request.

Let me know your decision.

larsbrinkhoff:
Note to self.

1. Fix wrapping around 1M memory.  NXM if writing to >= MAXMEM and MAXMEM <1M.
2. Add to user manual.

larsbrinkhoff:
I have yet to make the updates 1-3 above.  And I thought of one more thing.

larsbrinkhoff:
I pushed this aside to find the cause of the broken ITS build.  I'll let you know when I have made my updates.  I will squash the commits.

larsbrinkhoff:
Section to add to documentation:

### 3.3.3 Systems Concepts DC-10 Disk Controller (Device 610/614)

Systems Concepts DC-10 disk controller allowed IBM 2314 compatible disk drives to be attached to the MIT AI Lab PDP-10 running ITS.  This device is disabled by default.

larsbrinkhoff:
Hello @rcornwell,

I have updated, rebased, cleaned up, and dropped temporary commits.  Please merge.

rcornwell:
The following errors need to be fixed before this code can be merged.

`1>..\PDP10\ka10_ai.c(335) : warning C4244: 'function' : conversion from 'uint64' to 't_addr', possible loss of data
1>..\PDP10\ka10_ai.c(455) : warning C4018: '<' : signed/unsigned mismatch
1>..\PDP10\ka10_ai.c(561) : warning C4244: '=' : conversion from 't_offset' to 'long', possible loss of data
1>..\PDP10\ka10_ai.c(638) : warning C4018: '>' : signed/unsigned mismatch
1>..\PDP10\ka10_ai.c(699) : warning C4554: '<<' : check operator precedence for possible error; use parentheses to clarify precedence
1>..\PDP10\ka10_ai.c(998) : warning C4101: 'ctlr' : unreferenced local variable`

larsbrinkhoff:
I believe I fixed those warnings.

I noticed "makefile" doesn't pass much in the way of warning options to gcc.  I tested with -W -Wall and the compiler seems OK with ka10_ai.c.

larsbrinkhoff:
Changed to sim_fseeko: https://github.com/rcornwell/sims/compare/9bf6e9df0dd52d5803b904109b19ac730411a404..0756d77dd25238bebbc1f817ee619da5ae73dcd6

Still OK with ITS 724 and latest.

PR #136, opened 2019-08-05
KA10: aux CPU range should be configurable

larsbrinkhoff:
ITS 724 has the auxiliary CPU at 1,,040000 rather than 3,,000000.

The AUXCPURANGE define in kx10_cpu.c needs to be replaced with a run-time value determined by a CPU modifier.


larsbrinkhoff:
Pushed a commit.  Verified with ITS 724 and latest.

PR #142, opened 2019-08-10
SEL32: Instruction decode changes required by Gould diags

AZBevier:
Hosed this update, but should be OK with the changes.  I had a conflict with the SEL32 windows build, so I had to delete it.  I thought this was just a re-push, but must not be.  How to fix.


rcornwell:
I will try and merge this. However in the future please do all commit messages with SEL32:

AZBevier:
Rich,

I am sorry.  I guess I did not realize that the dialog that displays on a pull request is the same information I had entered with the commit.  I was using this as a message dialog to you, and not for messages for the commit.  I will not do that again.

All the machine types fail the Gould diag in the same place now.  It is in test 8 of the CV.TRP where the diag is trying to get a clock interrupt to fire on an RI instruction.  The level is already active, so it can not interrupt again.  Doing more testing I discovered the interrupt level had been active since the CV.CP2 diag.  I have not been able to resolve this error.  I am stuck for now.

On the UTX front, 32/27, 32/87, 32/97, and 32/67 (without Gould ECO??) will not run.  The Gould ECO seems to change the map protection bits like on the V6.  So it will not run a standard 32/67.  The V6 and V9 both panic in the startup initialization code in ioi_init() function where it calls the ioi_ictl() function to initialize system devices.  It handles the IOP OK.  It finds the DP at channel 800 and the SCFI at channel c00 and is successful at sending an INCH command and doing an ECI instruction to enable the channel interrupts for these disks.  It seems to die when trying to do a TIO on channel 0xfff.  Not sure why and what could be there at that address they are looking for.  Need to do more code search.

The interval timer code is changed to operate as required by UTX.  In MPX the clock is initialized at context switch time and started with the command to load, interrupt on zero and continue down counting.  It is used to balance the execution time amongst tasks.  In UTX, the clock is loaded and started and interrupts on zero.  The clock is to reload the original value, down count and interrupt on zero, and do it again until stopped.  I added the code to do that in sel32_clk.c, but you need to see if I did it correctly.  It is the option 0x39 to load & start, and 0x20 to stop.

Another device UTX is testing for is the MFP (Multi-Function Processor Model 8002) that is a replacement for the IOP.  It has two SCSI ports for disk and tape, RT clock, interval timer, and user interrupts.  May need to add this device in the future.

Tomorrow, Sunday, I will be on the road again for a couple weeks to Ruidoso, NM, so I will not be doing much on the simulator.  I will try to do some work on the DOC, as I can do that on my laptop.  We most likely will not have any internet connection where we will be staying, or cell phone for that matter.  I am sure we will have a good time without being ‘connected’.

Again, sorry for my confusion and have a great weekend.

Jim

PR #144, opened 2019-08-11
KA10: reading 7-track magtapes doesn't check against parity bit

larsbrinkhoff:
Computed parity from 6-bit frame isn't compared against the parity bit.

I looked at the CMP case and copied the code from there.


larsbrinkhoff:
With this fix in place, SALV 9 will TRAN from a 7-track tape.  It doesn't support 9-track tapes, so I had to go through this ordeal.

larsbrinkhoff:
Fixed by c26d5598429da319a3b07a566e88a5437b2176dd.

PR #145, opened 2019-08-13
KA10: ITS idling

larsbrinkhoff:
The typical ITS configuration runs at 100% host CPU utilization.  How do I fix this?  How do I know which devices are not idling properly?

### DIFF
--- a/PDP10/ka10_auxcpu.c
+++ b/PDP10/ka10_auxcpu.c
@@ -187,7 +187,7 @@ static t_stat auxcpu_svc (UNIT *uptr)
     auxcpu_ldsc.rcve = 1;
     uptr->wait = AUXCPU_POLL;
   }
-  sim_activate (uptr, uptr->wait);
+  sim_clock_coschedule (uptr, uptr->wait);
   return SCPE_OK;
 }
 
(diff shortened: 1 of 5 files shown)


larsbrinkhoff:
You can merge this or use the commit for reference.

rcornwell:
I can wait until you fix the TEN11 and AUXCPU. I think doing the merge will close the issue.

larsbrinkhoff:
Ok.

larsbrinkhoff:
I think I fixed it.  Tested with ITS 724.

I added UNIT_IDLE where missing, and used sim_clock_coscheule instead of sim_activate.  (I don't know what the difference between the two are, just that the former is magically better for idling.)

rcornwell:
sim_clock_coschedule does some magic to allow the device to go to sleep until some events occur. I did not want to touch the files to change over to that without you trying first. Just add everything to the commit and I will pull it in when I get home.

larsbrinkhoff:
Thanks.

I have added everything.

I tried to change DPY do coschedule, but it seems it broke the display.

rcornwell:
I don't think there is much we can do with DPY. If people want to allow ITS to IDLE, they should not enable that device.

larsbrinkhoff:
With coschedule, DPY idles well.  But the display doesn't work right, so it's not a solution.

I think we're done here for now.  All other ITS devices can do idling now.

PR #152, opened 2019-08-26
IMP DHCP is broken and does not follow the details spelled out in RFC2131 (Dynamic Host Configuration Protocol)

markpizz:
This was determined by the failure of a DHCP configured system to acquire 
an IP address.  That failure required WireShark examination of the protocol 
activities which then required review of RFC2131 to compare what was seen 
to what should have been seen.

What was seen:
1) First packet sent by the IMP DHCP client is a "DHCP Discover" to the 
   LAN broadcast address.  So far so good, maybe.  The details in the 
   "DHCP Discover" packet may or may not be correct.
2) 210ms later (before anything on the LAN gets a chance to respond), 
   the IMP DHCP client sends a "DHCP Request" packet.  This packet is 
   absolutely not proper according to the protocol.  It should be sent 
   in response to a DHCP Offer packet from the DHCP Server.  The RFC 
   says that "DHCP Discover" should be resent after some 4 seconds 
   (+/-1) with retries doubling each successive time until it exceeds 64 
   seconds.
   Note: On some networks (with switches in between the client and 
   the available DHCP Server) the network may take some time to 
   recognize the connectivity of the client system's sending MAC 
   address before responses can land at the right place, so some 
   retries for some clients may be completely normal.
3) 16+ seconds later another "DHCP Discover" packet is sent followed 
   20ms later by a "DHCP Request" packet.  Some approximately 16 
   seconds later another "DHCP Request" packet is sent which repeats
   indefinately (only "DHCP Request" packets) every 16 seconds.
   
It is somewhat surprising that no response came back from the DHCP 
server.  That may be due to one of two reasons:
1) The outgoing "DHCP Discover" was not properly formed.
OR
2) The broken protocol immediate sending of a "DHCP Request" packet
   may have quieted the DHCP server.
   
A review of the contents of the "DHCP Discover" packet show one 
difference between the IMP generated packet and a working node on 
the LAN.  This difference is the DHCP_OPTION_CLIENT_ID was not 
present.  It was added to the option list.  This didn't change 
anything since the "DHCP Request" packet is still being sent
some 200ms after the "DHCP Discover".

While looking for the logic that causes the "DHCP Request" to be 
sent, I encounter some strange scheduling activities.  It seems that
imp_unit[1] is dedicated to IMP timing activities, and it presumes that
sim_clock_coschedule() with an interval value of 1000 will somehow 
be useful to time against wall clock time and provide 1000 calls
per second.   The interval argument to sim_clock_coschedule() is 
in units of instructions and not anything related to wall clock.
This is quite a frequent calling expectation.

While walking through things I notice that sim_interval is being 
referenced as if it were the number of instructions that have been 
executed since the beginning of execution:

    case CONI:
        switch (GET_DTYPE(uptr->flags)) {
        case TYPE_MIT:
             last_coni = sim_interval;
AND
                 /* Only if there has been a CONI lately. */
                 if (last_coni - sim_interval < CONI_TIMEOUT)

The correct reference should be sim_gtime().  sim_interval has no 
meaning within any device implementation code.

The doc for the PDP10-KA describes the IMP, but doesn't mention 
MPX.  This has been clarified in the updated ka10_doc.doc file.

Looking through the MTAB entries there exists a SET IMP DHCPIP=a.b.c.d 
which doesn't actually have a parser for the IP address and manually 
setting the DHCP server would be meaningless.  This has been removed.
Additionally, displaying the DHCP server when DHCP is not enabled is 
meaningless, so the output of SHOW IMP DHCPIP has been changed to 
say that "DHCP disabled" when it is.

It is not clear why eth_filter() was called specifying that all 
multicast packets should be received.  Maybe so that explicit 
broadcast packets would be captured since implicitly they include
the broadcast address.  With all multi-cast packets enabled, specific
care must be taken to not misinterpret these packets.  It is better
to explicitly list the broadcast address as one of the potential 
addresses accepted in the eth_filter() API.  I disabled all multicast 
and we'll look at the broadcast reception issue later.  Disabling 
all multicast avoided some confusion interpreting some chatty 
traffic on the LAN and the quick sending of the DHCP REQUEST 
packets ceased..

Once the misinterpreted traffic was avoided, the imp_dhcp_timer() 
routine was observed looking at values which hadn't been set yet 
as a consequence of a successful DHCP negotiation (lease_time, etc).  
This is avoided by only examining that when the dhcp_state is set 
to BOUND.  Tracing traffic on the wire, it seems that the DHCP 
server is explicitly sending the DHCP Offer packet to the 
broadcast MAC address.  As a consequence, the call to eth_filter()
needed to be adjusted to specify both the IMP device MAC address 
and the broadcast address.

The protocol now moved along better, and it generated a DHCP Request
packet.  However, this packet was malformed.  There were multiple subtle
problems with both setting and interpreting data in the DHCP packets.
These were all related to referencing or providing things which were 
in network or host byte order consistently.  All IP addresses and 
masks should always be represented, presented and compared in network 
byte order.  Fixing this lets DHCP to work cleanly.

I created a separate unit to schedule wall clock activities against 
which are only used by DHCP retry activities and lease expiration 
and ARP aging.  This activity has been removed from the unit 1 IMP 
packet timer.

There was a problem with outbound connections when a NAT: attach was
made.  In this situation, the IMP traffic was on a simulated LAN within 
the sim_ether layer and traffic to nodes on the host system's LAN or
beyond all had to route through the sim_ether's NAT provided gateway.
The outbound IP packet flow would notice that this outbound traffic 
was not on the local LAN and therefore needed an ARP entry for the 
gateway.  An ARP request was made for the gateway while the original 
outgoing IP packet was queued for later delivery once an ARP entry 
was available.  Since the actual ARP request that was issued was 
specifically for the gateway IP address, no ARP replies would be 
coming in for the original packet's IP address.  This problem would 
be concealed if inbound traffic from the target IP happened first 
since that arriving traffic would create the needed ARP entry.  The 
solution to this problem was to generate a static ARP entry for the 
LAN gateway at attach time.  This then allows the outgoing traffic 
to off LAN destinations to avoid sending an ARP since it has already
determined that the packet must transit through the gateway and 
since the gateway ARP is persistently available, things now work.

Along the way to addressing the above problems, several latent problems 
were found in the sim_ether NAT code.  These were not related to the ARP 
problem, but were merely race conditions at NAT shutdown.  The race 
conditions appeared now since the IMP device's DHCP server sends a DHCP
RELEASE during the IMP detach.  All prior sim_ether use cases didn't send 
traffic while closing, so these hadn't been found previously.  
Additionally a problem with parsing of illegal NAT arguments were also
identified and fixed.

FYI, DHCP is enabled by default since the emulated IMP reflects normal 
behavior on a modern system.  Specifically, setting the IMP IP address or 
Gateway implicitly disables DHCP.  A SHOW IMP displays useful information 
about what is configured and/or happening.  A SHOW IMP ARP displays the 
current ARP information contained in the IMP device.

These are the cases which work now with the IMP device:

   1) Direct connection to the host system LAN interface:

		sim> set IMP enabled
		sim> set IMP mpx=4
		sim> set IMP host=10.3.0.6
	    sim> attach IMP eth0

	  This allows the simulated system (configured with IP address 
	  10.3.0.6) to reach systems on the LAN and across the Internet.  
	  The LAN IP address is acquired via DHCP.  The IMP internal
	  NAT will translate addresses from the 10.3.0.6 to the LAN's 
	  network based on DHCP acquired network parameters. On Windows 
	  hosts it also allows traffic to/from the host system.
	  
	  After the simulator is running, something like the following 
	  will be the output of a SHOW IMP command:
	  
		sim> SHOW IMP	  
		IMP     MAC=00:00:02:79:B8:A1, MPX=4, IP=192.168.60.40/24
				GW=192.168.60.6, HOST=10.3.0.6, DHCP Server IP=192.168.60.10, Lease Expires in 691159 seconds
				attached to eth0, DHCP, MIT
	  
   2) Direct connection to the host system LAN interface without 
      any NAT, presuming that the host system's network is 10.3.0.0/24
      and the gateway is 10.3.0.1:

		sim> set IMP enabled
		sim> set IMP mpx=4
		sim> set IMP GW=10.3.0.1
		sim> set IMP IP=10.3.0.6/24
	    sim> attach IMP eth0

	  This allows the simulated system (configured with IP address 
	  10.3.0.6) to reach systems on the LAN and across the Internet.  
	  On Windows hosts it also allows traffic to/from the host system.

	  After the simulator is running, something like the following 
	  will be the output of a SHOW IMP command:
	  
		sim> SHOW IMP	  
		IMP     MAC=00:00:02:79:B8:A1, MPX=4, IP=10.3.0.6/24
				GW=10.3.0.1, HOST=0.0.0.0, DHCP disabled
				attached to eth0, MIT

   3) Connectivity to the host system via NAT(slirp):

		sim> set IMP enabled
		sim> set IMP mpx=4
		sim> set IMP host=10.3.0.6
	    sim> attach IMP nat:tcp=2323:10.0.2.15:23,tcp=9595:10.0.2.15:95

	  This allows outbound traffic from the simulated system (configured
	  with IP address 10.3.0.6) to reach the LAN, the Internet and the 
	  host system.  The sim_ether NAT layer provides a DHCP server which 
	  will assign the address 10.0.2.15 to the simulated system and NAT 
	  translation will be performed by the sim_ether layer AND by the 
	  internal NAT in the IMP simulator.  The host (and external systems) 
	  can telnet into the simulator by telneting to the host system's 
	  IP address on port 2323.

	  After the simulator is running, something like the following 
	  will be the output of a SHOW IMP command:
	  
		sim> SHOW IMP	  
		IMP     MAC=00:00:02:79:B8:A1, MPX=4, IP=10.0.2.15/24
				GW=10.0.2.2, HOST=10.3.0.6, DHCP Server IP=10.0.2.2, Lease Expires in 86388 seconds
				attached to nat:tcp=2323:10.0.2.15:23,tcp=9595:10.0.2.15:95, DHCP, MIT

   4) Connectivity to the host system via NAT(slirp) using a single 
      layer of NATing:
	    
		sim> set IMP enabled
		sim> set IMP mpx=4
		sim> set IMP gw=10.3.0.1
		sim> set IMP IP=10.3.0.6/24
	    sim> attach IMP nat:nodhcp,gateway=10.3.0.1/24,tcp=2323:10.3.0.6:23,tcp=9595:10.3.0.6:95

	  This allows outbound traffic from the simulated system (configured
	  with IP address 10.3.0.6) to reach the LAN, the Internet and the 
	  host system.  The sim_ether NAT layer provides will be the only 
	  NAT translation.  The host (and external systems) can telnet into
	  the simulator by telneting to the host system's IP address on 
	  port 2323.

Connectivity beyond the above examples is also possible (using VDE or TAP, 
etc.).  These methods often require additional host specific network 
setup to talk to other systems.  VDE can be setup and which will allow the
simulator process to not run as root.


larsbrinkhoff:
Thank you, Mark!

CC @eswenson1, these are fixes for IMP using TAP.

markpizz:
@eswenson1 was the source of the test configurations that I used here.  His stuff let me focus on what I needed to do to find the issues.

They won't hurt a TAP based setup, but they are in no way specific to TAP.  

They will let a host system have the IMP connected directly to the LAN and if that system happens to have a dedicated interface for this role and a separate interface for normal host system communication to the LAN, then everyone will be happy.  Sure, you might say host systems with multiple interfaces would be rare, and that would be true for physical hosts, but, when the host happens to be running on a VM.  It also allows Windows based hosts to directly connect to the LAN and to talk to the host and everything else beyond.  Host to simulator communications can be easily achieved with a NAT connection, but reaching the simulator behind the IMP would then use non-standard port numbers.

Host to simulator setups can also be achieved with VDE.

At the end of the day, these fixes bring the IMP device up to being a complete user of sim_ether's LAN connectivity (with its features and foibles).

larsbrinkhoff:
I had the same alarming experience when I tried to experiment with Ethernet bridges on my VPS machine.  Eventually I cooked up the script below which seems to do what I need.  I'm not a networking expert either, so I may well have gotten some details wrong.

https://github.com/PDP-10/its/blob/master/build/pdp10-ka/tap.sh

larsbrinkhoff:
No, `make clean` and `git clean -xfd` will not clean the subrepositories.

eswenson1:
@markpizz You are correct that on my VPS with ITS under KLH10 and external access to/from ITS on this setup, I'm using iptables to do the IP forwarding and rules to specify which ports should be forwarded from outside to which ports on the ITS host.  

I'm perfectly happy doing the same thing with ITS under SIMH with tap -- if I could get it to work. So far, when I create a tap device, and let ITS (SIMH) connect to it, I'm unable to to get any traffic from my host to ITS (or the other way around).  If I could get this far, then I could use Iptables to forward inbound/outbound traffic to the Internet.  This is, of course, all that matters in the end. I don't care whether it is tun or tap, but I just need to be able to get ITS hosted on my Linux VPS without the need for a second NIC, or a second IP address, or a bridge (which won't work because I need a new MAC address that is acceptable by Linode, and they can't give me one -- I asked).

So if someone could (without a bridge) get ITS running under SIMH with tap, and figure out what magic needs to be done to allow local access from the host to ITS and vice-versa, then I'm sure I can get the iptables magic for external access -- just like I have for KLH10.

eswenson1:
I've tried this (most recently), with no success:
```
sudo tunctl -u eswenson -t tap0
sudo ifconfig tap0 10.3.0.1 up
echo 1 | sudo tee /proc/sys/net/ipv4/ip_forward
sudo iptables -A FORWARD -i tap0 -j ACCEPT
sudo iptables -A FORWARD -o tap0 -j ACCEPT
sudo iptables -t nat -A POSTROUTING -o eth0 -j MASQUERADE
```
And with that, I should be able to have traffic forwarded to the tap0 device, whose address is 10.3.0.1.  I then configure SIMH thus:
```
set imp enabled
set imp mac=2a:4b:23:2e:12:67
set imp mpx=4
set imp ip=10.3.0.1
set imp gw=66.175.218.1
set imp host=10.3.0.6
#set imp dhcp
at imp tap:tap0
```
Where GW is the actual gateway on eth0 that my host uses to get to the internet.
HOST is the IP of my ITS system.  And IP is the tap0 IP address.
When I bring up ITS, I cannot do "telnet 10.3.0.6" (or "supdup 10.3.0.6") from my Linux host.

markpizz:
Thanks for the example that doesn't work.  I do see some potential problems.

Meanwhile, what are all of the equivalent things for the KLH10 case that does work?  

Your case is more complicated than the simpleat example.

I'm really looking for the working example that I can map to the equivalent stuff with simh.

I think we are 6 timezones apart right now.  I'll be up by around noon EDT tomorrow, so anything you can provide by then will be helpful.

Ifconfig -a

As well as the output of the touring table on this Linux box.  I can't recall the command without poking around.

eswenson1:
Ok. I can make a tarball of a bootable ITS under KLH10 and provide you with the relevant config that allows connecting to ITS from the host. 

No special iptables setup is required. KLH10 creates the tun interface and configures everything itself. 

I do use iptables to allow forwarding and NAT from outside my host (Internet) and ITS, but that is really immaterial — if I can route local traffic from my host to ITS (and vice versa) then I can do the iptables magic to expose on the Internet.

The issue with SIMH for me is that I can’t even access ITS with tap from my host, so anything more complex isn’t going to work either.

eswenson1:
Oh, and to display the routes on Linux, you can use “route -n”, where the -n just specified to display addresses numerically, rather than trying to look up hostnames.

rcornwell:
One thing you might want to do is take my NATing code for FTP and add it into Slirp so that FTP will work under nat interface. I also have pretty fast code for updating the TCP checksums.

eswenson1:
I've placed a tarball of a bootable KLH10-based ITS system here:  https://s3.amazonaws.com/eswenson-its/public/mark-klh10.tgz.  

If you look in dskdmp.ini, you'll see the config and the IP address (192.168.1.100) for which the ITS is configured. 

To start it, type:  "sudo ./kn10-ks-its dskdmp.ini<enter>"

At the KLH10 prompt, type "go<enter>".  Then, you will end up in DSKDMP.  Type "its<enter>"  followed by "<escape>g".  ITS should start up.  Ignore the "top level interrupt" message you'll see. You should be able to "supdup 192.168.1.100" or "telnet 192.168.1.100" from your host.  No other setup should be required.

markpizz:
@eswenson1 
On your running KLH10 setup, will  you please provide the output of:
```
$ Ifconfig -a
$ route -n
```

eswenson1:
ifconfig -a:

```
dummy0    Link encap:Ethernet  HWaddr 5e:c9:16:ee:cb:8e
          BROADCAST NOARP  MTU:1500  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)

erspan0   Link encap:Ethernet  HWaddr 00:00:00:00:00:00
          BROADCAST MULTICAST  MTU:1450  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)

eth0      Link encap:Ethernet  HWaddr f2:3c:91:d4:29:2d
          inet addr:66.175.218.130  Bcast:66.175.218.255  Mask:255.255.255.0
          inet6 addr: 2600:3c01::f03c:91ff:fed4:292d/64 Scope:Global
          inet6 addr: fe80::f03c:91ff:fed4:292d/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:35488950 errors:0 dropped:0 overruns:0 frame:0
          TX packets:31368510 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:3163516197 (3.1 GB)  TX bytes:130838388457 (130.8 GB)

eth0:1    Link encap:Ethernet  HWaddr f2:3c:91:d4:29:2d
          inet addr:192.168.219.106  Bcast:0.0.0.0  Mask:255.255.128.0
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1

gre0      Link encap:UNSPEC  HWaddr 00-00-00-00-80-00-00-00-00-00-00-00-00-00-00-00
          NOARP  MTU:1476  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)

gretap0   Link encap:Ethernet  HWaddr 00:00:00:00:00:00
          BROADCAST MULTICAST  MTU:1462  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)

ip6_vti0  Link encap:UNSPEC  HWaddr 00-00-00-00-00-00-00-00-00-00-00-00-00-00-00-00
          NOARP  MTU:1364  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)

ip6gre0   Link encap:UNSPEC  HWaddr 00-00-00-00-00-00-00-00-00-00-00-00-00-00-00-00
          NOARP  MTU:1448  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)

ip6tnl0   Link encap:UNSPEC  HWaddr 00-00-00-00-00-00-00-00-00-00-00-00-00-00-00-00
          NOARP  MTU:1452  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)

ip_vti0   Link encap:IPIP Tunnel  HWaddr
          NOARP  MTU:1480  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)

lo        Link encap:Local Loopback
          inet addr:127.0.0.1  Mask:255.0.0.0
          inet6 addr: ::1/128 Scope:Host
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
          RX packets:760866 errors:0 dropped:0 overruns:0 frame:0
          TX packets:760866 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:43559858 (43.5 MB)  TX bytes:43559858 (43.5 MB)

sit0      Link encap:IPv6-in-IPv4
          NOARP  MTU:1480  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)

tap0      Link encap:Ethernet  HWaddr aa:5f:40:24:d7:77
          inet addr:10.3.0.1  Bcast:10.255.255.255  Mask:255.0.0.0
          inet6 addr: fe80::a85f:40ff:fe24:d777/64 Scope:Link
          UP BROADCAST MULTICAST  MTU:1500  Metric:1
          RX packets:127 errors:0 dropped:20 overruns:0 frame:0
          TX packets:70 errors:0 dropped:2 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:18585 (18.5 KB)  TX bytes:5072 (5.0 KB)

teql0     Link encap:UNSPEC  HWaddr 00-00-00-00-00-00-00-00-00-00-00-00-00-00-00-00
          NOARP  MTU:1500  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:100
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)

tun0      Link encap:UNSPEC  HWaddr 00-00-00-00-00-00-00-00-00-00-00-00-00-00-00-00
          inet addr:66.175.218.130  P-t-P:192.168.1.100  Mask:255.255.255.255
          inet6 addr: fe80::7027:cbb:a73c:c91e/64 Scope:Link
          UP POINTOPOINT RUNNING NOARP MULTICAST  MTU:1500  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:1 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:500
          RX bytes:0 (0.0 B)  TX bytes:48 (48.0 B)

tunl0     Link encap:IPIP Tunnel  HWaddr
          NOARP  MTU:1480  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)
```

route -n:

```
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         66.175.218.1    0.0.0.0         UG    0      0        0 eth0
0.0.0.0         66.175.218.1    0.0.0.0         UG    0      0        0 eth0
10.0.0.0        10.3.0.1        255.0.0.0       UG    0      0        0 tap0
66.175.218.0    0.0.0.0         255.255.255.0   U     0      0        0 eth0
192.168.1.100   0.0.0.0         255.255.255.255 UH    0      0        0 tun0
192.168.128.0   0.0.0.0         255.255.128.0   U     0      0        0 eth0
```

larsbrinkhoff:
I'm updating my "TT" ITS to use the latest pdp10-ka.  I will try to have it use the TAP interface.

PR #157, opened 2019-09-04
KA10: Fix IMP ICMP and add detailed debug options

markpizz:
- ICMP Echo and Echo reply packets mangled the ICMP checksum when no
   mangling was needed.
- Other ICMP message types may still be wrong since they potentially contain
   IP header information that may or may not need IP addresses manipulated
- Increased the ARP table size from 8 to 64 since the ARP RFC (826) suggests
   that ARP info from all received ARP packets be gathered as ARP broadcasts
   are received.  A normal LAN can easily blow over an 8 entry ARP table.

### DIFF
--- a/PDP10/kx10_imp.c
+++ b/PDP10/kx10_imp.c
@@ -145,7 +145,7 @@
 #endif
 #endif
 
-#define IMP_ARPTAB_SIZE        8
+#define IMP_ARPTAB_SIZE        64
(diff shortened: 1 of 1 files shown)


rcornwell:
Looks good. I like the debug logging methods. Will make it easier to debug problems in future. Will pull this in tonight.

Issue #153, opened 2019-08-26
KA10: Terminal multiplexers should not write to txdone


larsbrinkhoff:
@markpizz pointed out txdone is private to TMXR.

larsbrinkhoff:
Also don't set UNIT_DISABLE.

larsbrinkhoff:
Also check why DPK only allows one connection.

larsbrinkhoff:
Also this:

> In order for the TMXR layer to know about which units manage which
> lines, the device reset routine should call:
>
> t_stat tmxr_set_line_unit (TMXR *mp, int line, UNIT *uptr_poll);
> t_stat tmxr_set_line_output_unit (TMXR *mp, int line, UNIT *uptr_poll);

larsbrinkhoff:
I made the requested fixes to the MTY device.  It seems to work exactly like before, no better or no worse.  Still sluggish compared to using sim_activate.

larsbrinkhoff:
The changes are on my branch here:  
https://github.com/larsbrinkhoff/ka10-simh/commits/lars/tmxr

rcornwell:
If this works, push it to the repo. I am planing on pushing the current system to the main simH tree within the next day or two.

larsbrinkhoff:
I'll try.  If I miss the window, no big deal.

larsbrinkhoff:
I posted a pull request: #159.  I still have to fix DPK to allow more than one connection.

larsbrinkhoff:
Based on input from @markpizz, I'm reconsidering the implementation of MTY.  The other two, TK10 and DPK should be OK.

I'm thinking something along the line of two UNITs for the DEVICE.  One unit polls for connections and input.  The other is for output.  The output unit should keep track of one word being output for each line, and feed data one character at a time to tmxr_putc_ln.

larsbrinkhoff:
Oh, and the output unit could schedule with sim_activate for speedy service, in case sim_clock_coschedule remains sluggish.  It need not be active when all lines are idle.

larsbrinkhoff:
I'll close this one, because the txdone thing has been fixed.

markpizz:
> I'm thinking something along the line of two UNITs for the DEVICE. One unit polls for connections and input. The other is for output. The output unit should keep track of one word being output for each line, and feed data one character at a time to tmxr_putc_ln.

[...]

PR #161, opened 2019-09-10
KA10: Restructure MTY device.

larsbrinkhoff:
Along the lines in https://github.com/rcornwell/sims/issues/153#issuecomment-529102686

### DIFF
--- a/PDP10/ka10_mty.c
+++ b/PDP10/ka10_mty.c
@@ -1,6 +1,6 @@
 /* ka10_tk10.c: MTY, Morton multiplex box: Terminal multiplexor.
 
-   Copyright (c) 2018, Lars Brinkhoff
+   Copyright (c) 2018-2019, Lars Brinkhoff
 
    Permission is hereby granted, free of charge, to any person obtaining a
    copy of this software and associated documentation files (the "Software"),
@@ -37,6 +37,7 @@
 #define MTY_NAME        "MTY"
 #define MTY_DEVNUM      0400
 #define MTY_LINES       32
+#define MTY_FIRST       C1      /* Frist character in output word. */
 
 #define MTY_PIA         0000007 /* PI channel assignment */
 #define MTY_RQINT       0000010 /* Request interrupt. */
@@ -50,7 +51,8 @@
 #define MTY_CONO_BITS   (MTY_PIA | MTY_LINE)
 
 static t_stat       mty_devio(uint32 dev, uint64 *data);
-static t_stat       mty_svc (UNIT *uptr);
+static t_stat       mty_input_svc (UNIT *uptr);
+static t_stat       mty_output_svc (UNIT *uptr);
 static t_stat       mty_reset (DEVICE *dptr);
 static t_stat       mty_attach (UNIT *uptr, CONST char *cptr);
 static t_stat       mty_detach (UNIT *uptr);
@@ -59,13 +61,18 @@ static t_stat       mty_help (FILE *st, DEVICE *dptr, UNIT *uptr,
                                int32 flag, const char *cptr);
 extern int32        tmxr_poll;
 
+static uint32       mty_active_bitmask;
+static uint64       mty_output_word[MTY_LINES];
+static int32        mty_input_character;
+
 TMLN mty_ldsc[MTY_LINES] = { 0 };
 TMXR mty_desc = { MTY_LINES, 0, 0, mty_ldsc };
 
 static uint64 status = 0;
 
 UNIT                mty_unit[] = {
-    {UDATA(mty_svc, TT_MODE_7B|UNIT_IDLE|UNIT_ATTABLE, 0)},  /* 0 */
+    {UDATA(mty_input_svc, TT_MODE_7B|UNIT_IDLE|UNIT_ATTABLE, 0)},  /* 0 */
+    {UDATA(mty_output_svc, UNIT_DIS|UNIT_IDLE, 0)},  /* 0 */
 };
 DIB mty_dib = {MTY_DEVNUM, 1, &mty_devio, NULL};
 
@@ -85,7 +92,7 @@ MTAB mty_mod[] = {
 
 DEVICE mty_dev = {
     MTY_NAME, mty_unit, NULL, mty_mod,
-    1, 8, 0, 1, 8, 36,
+    2, 8, 0, 1, 8, 36,
     NULL, NULL, mty_reset, NULL, mty_attach, mty_detach,
     &mty_dib, DEV_DISABLE | DEV_DIS | DEV_DEBUG, 0, dev_debug,
     NULL, NULL, mty_help, NULL, NULL, mty_description
@@ -97,7 +104,6 @@ static t_stat mty_devio(uint32 dev, uint64 *data)
     TMLN *lp;
     int line;
     uint64 word;
-    int ch;
 
     switch(dev & 07) {
     case CONO:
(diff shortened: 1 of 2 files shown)


larsbrinkhoff:
This is now both good with idling and has fast output.  Much help from @markpizz.

rcornwell:
Will pull this over in a bit... need to try and get performance working on KL10. And test PD changes under KA.

larsbrinkhoff:
Sounds fine.

I don't think the KA10 version of the PD clock had a clock interrupt.
So either make it `#if KL_ITS`, or make it settable `SET PD INTERRUPT`
or something.

eswenson1:
By the way, @markpizz, to answer a question you asked, the ITS I gave you did have MTY device enabled.  If you look in the "run" file, you will see:
```
set mty enabled
at mty 10003
```
So once you have your ITS booted, you can "telnet localhost 10003" to get to the MTY device.  Type ^Z there in order to get a HACTRN and DDT.

markpizz:
@eswenson1 Thanks.

I had seen that the device was enabled in the configuration file, but on a simulator today, at least for me, it is very easy enable a device, but far more complicated to configure an operating system to actually enable the use of that device.  

My uncertainty was confirmed when I initially got no response on the telnet session connected to the MTY device.  When I dug a little deeper, I realized that that the OS must actually have some awareness of the device since I saw DATAO activity to this device.  Therefore, given that the OS was trying to output data, I could then dig into what was wrong with the device emulation.

eswenson1:
@markplzz I’m sorry I wasn’t complete in my response. Not only was the MTY device configured in the SIMH script (run), but it was configured in ITS. The image I gave you was able to support “telnet localhost 10003”. I do that regularly. After telnetting to the configured port, however, you have to type Control-Z to get ITS to create a session and start reading and writing to the terminal.

markpizz:
@eswenson1 I figured that out since, as I said, I saw DATAO activity on that device.  Once I knew I had a good test case I was able to observe the issue and provide something that actually worked.

Thanks.

markpizz:
@larsbrinkhoff I just looked at the current code.

In the output service routine you've got:
```
    for (i = 0; i < MTY_LINES; i++) {
        /* Round robin scan 32 lines. */
        scan = (scan + 1) & 037;

        if ((mty_active_bitmask & (1 << scan)) == 0 ||
            (txdone = tmxr_txdone_ln (&mty_ldsc[scan])) == 0)
            continue;

        /* Write up to five characters extracted from a word.  NUL
           can only be in the first character. */
        word = mty_output_word[scan];
        if (word != 0) {
            ch = (word >> 29) & 0177;
            ch = sim_tt_outcvt(ch, TT_GET_MODE (mty_unit[0].flags));
            if (tmxr_putc_ln (&mty_ldsc[scan], ch) != SCPE_STALL)
                mty_output_word[scan] = (word << 7) & FMASK;
        } else if (txdone == 1) {
            sim_debug(DEBUG_DETAIL, &mty_dev, "Output ready line %d\n", scan);
            status &= ~MTY_LINE;
            status |= scan << 12;
            status |= MTY_ODONE;
            set_interrupt(MTY_DEVNUM, status & MTY_PIA);
            mty_active_bitmask &= ~(1 << scan);
        }
    }
```
This almost looks good.  By almost, I think it needs a "break;" after the line which clears the mty_active_bitmask bit for the scanned line.

This is needed to address the case where multiple lines are active concurrently (something you haven't tested with one connection), to avoid overwriting the "status" register which indicates the line that an interrupt was generated for.  This suggestion follows the similar logic in the input service routine where you stop scanning when you've found a line which has input pending.  If I'm reading this wrong, please explain.

Also, looking closely at the input service routine, you are explicitly setting mty_ldsc[i].conn = 1;
the conn field is open for reference by you, but not to be directly changed.  In that particular place, you are changing it to what it to the value it already contains, so that line is harmless, but really shouldn't be there anyway.  You really shouldn't need to reference the conn field anyway (which is done in the later scan loop) since the internals of TMXR will behave reasonably for both connected and unconnected lines.

The input path data path processing isn't quite right.  The problem is that you're using tmxr_input_pending_ln() to determine if an input characters is ready for reading.  That API is really meant for internal use by the TMXR layer.  Your 3 devices (MTY, DPK and TK10) are the only non simh internal devices that have ever used it.  The public API to test for and fetch an input character is tmxr_getc_ln().  When it returns a character (with the TMXR_VALID bit set), you've got something to process.  That result should be buffered and an incoming character interrupt should be generated.  The DATAI logic should take the received character and clear the fact that an incoming character is pending.  The polling logic scan should avoid scanning for more input on this (or any) line until the received character has been received by the simulator in the DATAI path.  The DATAI path should specifically call sim_activate_abs(unit[0], 0) after the received character has been read by the OS.

markpizz:
FYI: The doc/simh.doc section that describes the TMXR APIs specifically does not mention the tmxr_input_pending_ln() API.

larsbrinkhoff:
Thanks Mark, I updated the commit.

larsbrinkhoff:
I added on a somewhat unrelated commit.  The TEN11 device has been very slow to start responding since I changed it to use sim_clock_coschedule rather than sim_activate.  Once it's up an running, it's OK.

I don't know why that is, but anyway making it start the svc routine right away fixes the problem.

rcornwell:
Let me know when you get Marks changes in place and tested. I will hold off on pulling this in until it is good.

larsbrinkhoff:
These are Mark's changes, along with mine, and changes suggested by Mark.  It's all in there.

I tested it with two simultaneous connections both writing a file at full speed.

markpizz:
Almost there.  Two very minor changes.

In the MTY DEVICE declaration, the number of units specified should be 2 instead of 1, AND the output unit's flag bit initializer should include UNIT_DIS.

----------
I see you've made changes to the PD clock.  Looking through that code, it seems that there are programmatic conditions that turn this clock on and/or off.  That's absolutely fine.  However, if that is going on (or possible), then this device isn't a simh calibrated clock.  The tick interval it programs is already correct, so there is no problem there.  Just remove these lines:
```
    int32 t;

    t = sim_rtcn_calb (pd_tps, TMR_PD);
```

larsbrinkhoff:
Thanks Mark, I added in those changes.

Maybe you clicked the "force-pushed" link to see my updates, and then you saw the PD device changes.  Those are because I rebased my branch on top of Rich's latest master.  So, @rcornwell, those comments are really directed at you.

markpizz:
Ahh, yes.  That was Rich's change.

rcornwell:
@markpizz The clock does not really turn off . It is just providing a 60hz interrupt. But I can remove the line, since I don't use the result.

markpizz:
I see programmable logic which starts and stops the potentially consistent scheduling of the service routine in the CONO code path and the service routine.  I'm sure that the device works correctly, but having a device which optionally goes on and off is not a simh calibrated clock.  Invoking some of those APIs can interfere with precise clock calibration.

Meanwhile, I see that the PD device has the UNIT_DISABLE flag set in the UNIT flags, this shouldn't be. 
Probably the only devices that should ever have UNIT_DISABLE are those which explicitly have UNIT_ATTABLE and are most likely disk or tape devices.  If a unit exists only for the purposes of scheduling some otherwise orthogonal events (the the MTY output unit), it should have the UNIT_DIS flag set.  DEVICEs which can optionally be enabled should have the DEVICE_DISABLE bit set in the DEVICE flags.  @rcornwell , there are many examples of UNIT_DISABLE being incorrectly used in your simulators.

rcornwell:
I will have to check which units have UNIT_DISABLE in them that should not. Devices are disabled so as to not enter into the I/O space. Any device which has DEV_DIS will  not be added to I/O Space. I was under the impression that if a UNIT did not have UNIT_DISABLE flag that it could not be enabled/disabled. Some of my devices are internal devices which should not be changed.

markpizz:
> I was under the impression that if a UNIT did not have UNIT_DISABLE flag that it could not be enabled/disabled. Some of my devices are internal devices which should not be changed.

That impression is true, BUT it specifically makes sense only for multi-unit device that have UNITs that correspond 1-to-1 to things like disk drives connected to a disk controller DEVICE or tape drives connected to a tape controller DEVICE.  

Places where UNIT_DISABLE doesn't belong:
b5500_dtc.c
b5500_urec.c
i7000_cdp.c
i7000_cdr.c
i7000_chron.c
i7000_lpr.c
i7010_chan.c
i7090_drum.c
i7090_hdrum.c
i7090_lpr.c
...
ka10_imx.c
ka10_pd.c
ka10_skt.c
ka10_cp.c
kx10_cr.c
kx10_dc.c

Places where UNIT_DISABLE might belong:
b5500_io.c
i701_chan.c
i7010_chan.c
i7070_chan.c
i7080_chan.c
i7090_chan.c

rcornwell:
I can understand why for some of these. However the Card Punch/Card readers which support more then one unit, why is it wrong to be able to disable individual units?

b5500_dtc.c - probably does not need to be able to disable unit.
b5500_urec.c - Why can't individual units be disabled?

i7000_* - Either individual devices, or in some cases, disabling the device rather then the controller.

markpizz:
> b5500_urec.c - Why can't individual units be disabled?

Well, there doesn't seem to be anything in the code path that would behave differently one way or another if the units were disabled especially the output units.

Apart from this, any DEVICE with only 1 unit should have a disable on the DEVICE and not the unit.

rcornwell:
I agree, however some devices have 1 unit on some systems, and more then one unit on other systems. (I7000 here). I can audit this. I recommend creating a new issue if you want it fixed.

PR #164, opened 2019-09-17
KA10: Fix inconsistent address types

markpizz:
auxcpu_base was previously defined as a t_value, while it should have
been a t_addr everywhere.  This then showed where it was referenced
need declaration adjustments to also work with t_addr in expressions.

### DIFF
--- a/PDP10/ka10_auxcpu.c
+++ b/PDP10/ka10_auxcpu.c
@@ -54,7 +54,7 @@
 
 static int pia = 0;
 static int status = 0;
-t_value auxcpu_base = 03000000;
+t_addr auxcpu_base = 03000000;
 
 static t_stat auxcpu_devio(uint32 dev, t_uint64 *data);
 static t_stat auxcpu_svc (UNIT *uptr);
@@ -407,13 +407,13 @@ static t_stat auxcpu_set_base (UNIT *uptr, int32 val, CONST char *cptr, void *de
     if (r != SCPE_OK)
         return SCPE_ARG;
 
-    auxcpu_base = x;
+    auxcpu_base = (t_addr)x;
     return SCPE_OK;
 }
 
 static t_stat auxcpu_show_base (FILE *st, UNIT *uptr, int32 val, CONST void *desc)
 {
-    fprintf (st, "Base: %llo", auxcpu_base);
+    fprintf (st, "Base: %" T_ADDR_FMT "o", auxcpu_base);
     return SCPE_OK;
 }
 #endif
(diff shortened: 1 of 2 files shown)


rcornwell:
Be careful changing the type of addr in Mem_read_its/Mem_write_its, this is used as an index and passed to page_lookup_its.

markpizz:
No doubt.

However, it is an address, and in this simulator a t_addr is a uint32 which is certainly reasonable for this system, right?  The logic in those procedures compares the supplied values to various other t_addr values such as CPU.capac.

rcornwell:
Mem_read/Mem_write function on the KA are pointers that are set in the initialization routine.

larsbrinkhoff:
The auxcpu base address is a physical address which can be 20 bits.  Even on a KI/KL the maximum is 22 bits, right?  So a 32-bit t_addr should be OK.

Extended addressing could potentially use 30-bit addresses (just 23 on KL10 model B), or 36 if you count byte pointers.

markpizz:
@rcornwell I see what you're talking about.  The Mem_read and Mem_write routines are different for the PDP6 vs the other systems.  A similar change is needed for the PDP6 implementations.

The initial problem was noticed by a link time warning issued on the AppVeyor Release build of the code on Windows.  The PDP10-KA is built near the end of the things.

rcornwell:
The changes also need to be made to all the page_lookup functions. Also don't forget the KL10 branch which I plan to update tonight.

larsbrinkhoff:
Did this show up in Visual Studio?

markpizz:
> The changes also need to be made to all the page_lookup functions. Also don't forget the KL10 branch which I plan to update tonight.

Well, that seems to be a much more pervasive change.  The changes I've made already were simple and bounded to a couple of places.  You are currently using explicit uint32 declarations for AB and PC where the normal simh convention for things like this would be to use a t_addr type.  Look at:

t_stat fprint_sym (FILE *of, t_addr addr, t_value *val, UNIT *uptr, int32 sw);
t_stat parse_sym (CONST char *cptr, t_addr addr, UNIT *uptr, t_value *val, int32 sw);

I think it's best for you to review these things, right?

> Did this show up in Visual Studio?

Yes.

My normal build is for a Debug build which didn't show this.  The last time I did a Release build I didn't review every line in the build log?

markpizz:
It is unless USE_INT64 && USE_ADDR_64 are both true.  Then t_addr is a uint64

rcornwell:
Ok, I can try and go through and switch addresses to t_addr.

rcornwell:
I have changed this in the KL10 branch.

rcornwell:
These changes were done in the KL10 branch, since this is merged this request can be closed.

PR #179, opened 2020-01-30
KA10: SDL joystick support

larsbrinkhoff:
Posting a draft to add support for SDL joysticks and game controllers.

@markpizz, please review the sim_video part.


markpizz:
Hi Lars,

Can you explain, in abstract concepts, how or what this represents (and how it maps to things that have come before).  

Is the Joystick support merely a way to move a logical cursor around, and if so, how does that logical cursor movement get represented into simulated devices?  Is this motion activity presented as mouse activity in a simulator?
Are Game controller button activities presented as keyboard events?

Thanks.

- Mark

larsbrinkhoff:
@markpizz,

In this pull request, the joysticks or game pads map to Spacewar "consoles" (the word they used for game controllers).  There's no cursor, mouse, or keyboard involved.

Pushing the stick left and right makes the ships rotate.  Pushing forward adds thrust.  Pushing a button fires a torpedo.

larsbrinkhoff:
@PiDP, here is draft support for USB joysticks & gamepads.  I humbly suggest these are more fun to use than a keyboard.  Especially if there are four players!

This code assumes the first two axes are the "primary stick".  This seems to hold for most USB devices.  The first axis is rotate left/right, and the second is thrust.  The buttons are, in order, fire, hyperspace, and beacon.  Beacon is only available with the OCNSLS device used by WAR.

larsbrinkhoff:
@markpizz, here's another consideration.

I'd like to make the host gaming devices available for many different guest games.  There are two versions of Spacewar using two different hardware input devices.  And there's a PDP-10 version of Dazzle Dart that uses a mix of inputs, among those a 128-channel A/D device.

So I'd like the user to SET ENABLE one or more of those devices as appropriate.  As you can see in the commit, I modeled my change on your vid_display_kb function pointer callback.  But maybe it would be better to call a function to register a callback.  Then multiple devices could access the gaming device events.

markpizz:
Why do you have a separate vid_init_controllers()?  Why wouldn't the things being done in vid_init_controllers() merely be part of what is normally a consequence of a call to vid_open() which happens when display.c calls ws_init().

> So I'd like the user to SET ENABLE one or more of those devices as appropriate. As you can see in the commit, I modeled my change on your vid_display_kb function pointer callback. But maybe it would be better to call a function to register a callback. Then multiple devices could access the gaming device events.

I think the callback model is the right approach.

You could certainly add a register function, which would then take each of the possible callback functions as an argument, and that function should only return success if the particular callback it want that's trying to be registered hasn't already been registered.    The vid_close() function should clear any registered callbacks, vid_close() should be called directly (or indirectly) as a consequence of the specific video device being disabled..

larsbrinkhoff:
`vid_init_controllers` is just an artifact of my experiments.  `vid_open` sounds great.

I'll update the commit as per your instructions.

larsbrinkhoff:
@markpizz, @rcornwell 

Updated.  Comments?

markpizz:
1) I don't think the #if expression that defines HAVE_GAMECONTROLLERFROMINSTANCEID actually reflects the version information in the comments right there.  Fix the comments or the logic.
2) In sim_video.c/vid_video_events(), where you're calling your new vid_joy_buffon, etc. event handling routines, you're passing a pointer to the event union member (which is certainly valid), however this is not what the other event handling functions that are already there are passing.  They pass a pointer to the generic event union and cast the pointer to the desired event pointer type.  You should change either the old ones to do the new thing or the new ones to do the old thing.
3) In your conditional depending on HAVE_GAMECONTROLLERFROMINSTANCEID at line 589, it looks like the ending 'else' in that conditional block should have the opening "{" character from line 599.
4) Any registration callback calls should be done in the respective device attach routines, after a successful vid_open.  The calls to vid_register_quit_callback() should be moved to the attach routines.  The vid_close should also clear the quit callback.

larsbrinkhoff:
Thanks.  Regarding point 4, the KA10 DPY device uses the display.h API which hides vid_open.  And I don't think there's any call to vid_close.

markpizz:
No dispute on point 4, but vid_open must be called inside of display_init(), so my point about the callback registrations should occur after display_init() since you're clearing the registrations in vid_open().

Maybe its time to extend the set of routines provided in the display code which provides the equivalent of closing the device.  Since the devices that use display don't tend to have attach routines, the new display_close(or whatever it gets named) should be called from the reset routine when the device status transitions from enabled to disabled.

larsbrinkhoff:
1. Updated comment.
2. Made pointers consistent.
3. If I moved the `{` inside the preprocessor conditional, it would be unbalanced if HAVE_GAMECONTROLLERFROMINSTANCEID is false.  If this logic and/or indentation is too messy, I could put HAVE_GAMEC... in the if() instead.
4. Added display_close.

larsbrinkhoff:
The call to display_init is from ty340_reset.

markpizz:
> 3. If I moved the { inside the preprocessor conditional, it would be unbalanced if HAVE_GAMECONTROLLERFROMINSTANCEID is false. If this logic and/or indentation is too messy, I could put HAVE_GAMEC... in the if() instead.

Part of the problem is that the indentation convention you're using is not what is used elsewhere in this module.

Specifically, the opening { generally follows the if or else, and the closing } is indented as much as the lines it encloses.  Both the if and any related else that may follow are indented the same amount.  For example:

```
    if (banana) {
        eat-an-apple;
        make-some-banana-bread;
        }
    else {
        have-ice-cream;
        take-a-nap;
        }
```

Please follow the indentation convention already in place in the module that you're modifying.  This is the simh project standard.  Folks creating new modules get to define the convention they want to use in the modules they author (presuming it is used consistently).  Folks modifying existing modules need to follow the pre-existing convention in the module being modified so that someone coming along later doesn't get confused by inconsistencies.

After you fix the brace indentation I think you'll see the point I raised.

sim_video.c
```diff
@@ -510,6 +548,71 @@ return SCPE_OK;
...
+        if (SDL_IsGameController (i)) {
+            SDL_GameController *x = SDL_GameControllerOpen (i);
+            if (x != NULL) {
+                sim_debug (SIM_VID_DBG_VIDEO, vid_dev,
+                "Game controller: %s\n", SDL_GameControllerNameForIndex(i));
+                }
+            }
+        else
+#endif
+             {
```
larsbrinkhoff: ???

larsbrinkhoff:
I have no problem following a coding convention.  I have tried to update the commit as per your example.

Sorry, I still don't understand what you mean about the `else` and the `{`.  Obviously it looks very awkward now.  I can't make it `else {` without changing something else too.

markpizz:
Hmm...  So just to test I changed the #define for HAVE_GAMECONTROLLERFROMINSTANCEID to be 0.

It then compiled without error.  So, puzzled, I studied the brace nesting and finally saw:
```
    for (i = 0; i < n; i++) {
             {
            y = SDL_JoystickOpen (i);
            if (y != NULL) {
                [...]
                }
            }
        }
```
The extra nested braces is the visual and indentation problem that confused me (and possibly someone else who comes later).

How about:
```
    for (i = 0; i < n; i++) {
#if HAVE_GAMECONTROLLERFROMINSTANCEID
        if (SDL_IsGameController (i)) {
            SDL_GameController *x = SDL_GameControllerOpen (i);
            if (x != NULL) {
                sim_debug (SIM_VID_DBG_VIDEO, vid_dev,
                "Game controller: %s\n", SDL_GameControllerNameForIndex(i));
                }
            }
        else {
#else
        if (1) {
#endif
            y = SDL_JoystickOpen (i);
            if (y != NULL) {
                sim_debug (SIM_VID_DBG_VIDEO, vid_dev,
                "Joystick: %s\n", SDL_JoystickNameForIndex(i));
                sim_debug (SIM_VID_DBG_VIDEO, vid_dev,
                "Number of axes: %d, buttons: %d\n",
                SDL_JoystickNumAxes(y),
                SDL_JoystickNumButtons(y));
                }
            }
        }
```
Which is logically equivalent and simpler (at least for me) to read.  

Or, since this HAVE_GAMECONTROLLERFROMINSTANCE stuff is working around a deficiency  under the covers in older versions of SDL, rather than a missing API, maybe this:

```
    for (i = 0; i < n; i++) {
        if (SDL_IsGameController (i)) {
            SDL_GameController *x = SDL_GameControllerOpen (i);
            if ((x != NULL) && HAVE_GAMECONTROLLERFROMINSTANCEID) {
                sim_debug (SIM_VID_DBG_VIDEO, vid_dev,
                "Game controller: %s\n", SDL_GameControllerNameForIndex(i));
                }
            }
        else {
            y = SDL_JoystickOpen (i);
            if (y != NULL) {
                sim_debug (SIM_VID_DBG_VIDEO, vid_dev,
                "Joystick: %s\n", SDL_JoystickNameForIndex(i));
                sim_debug (SIM_VID_DBG_VIDEO, vid_dev,
                "Number of axes: %d, buttons: %d\n",BOM
                SDL_JoystickNumAxes(y),
                SDL_JoystickNumButtons(y));
                }
            }
        }
```
Which I think is equivalent.

Now, thinking more about the fact that this is an under the covers version issue with SDL2, I'm not sure that this should be a compile time check.  On many platforms the SDL2 shared object will be how the SDL2 code is accessed and it may readily change independent of when or how a simulator is compiled....  You could easily use SDL_GetVersion() to determine how you would then compute a global variable reflecting what HAVE_GAMECONTROLLERFROMINSTANCE does.

--------

Apart from this, your commit comments mentioning changing the PDP10, while it really should be KA10 and KI10.  The PDP10 simulator is the KS10.

larsbrinkhoff:
I think I need to check both at compile time (or else we might try to call an undeclared function), and at run time.

larsbrinkhoff:
Updated:
- Added runtime version check.
- Used `if (1)` for increased readability.
- Changed commit comments to KA10.

Note that HAVE_GAMECONTROLLERFROMINSTANCEID also guards against SDL1 which (as far as I understand) doesn't have the game controller API.

markpizz:
Don't worry about SDL1.  I haven't had a chance to rip out that code.   We initially were forced to support both SDL1 and SDL2 since when the video functionality was first added SDL2 hadn't seen wide deployment yet so we needed to support both.  That has long past.   Supporting this functionality is really only targeted for relatively modern host systems.  All of them readily have SDL2 support.

larsbrinkhoff:
In that case I can reduce the number of #if's.

markpizz:
Looks good now.

Please flow this sequence of commits into the github.com/simh/simh

larsbrinkhoff:
It's not so easy with the fork structure.  I can submit a pull request with the commit tagged "video".  Then I can add the PDP-1 and PDP-11 things on top later.

The KA10 commits are @rcornwell's ownership.

markpizz:
The "fork" structure really isn't a fork, but a copy that passes things back and forth.  Harder to manage than true repo forks,

In any case, you've already started with the video changes.  Don't forget the display.c change.  The simulator specific changes can come after these.

larsbrinkhoff:
@rcornwell, Mark has approved the commits tagged "video" and "display".  Do you have any comments on the "KA10" commits?

rcornwell:
They all look good to me. Let me finish off the NIA20 so that It can at least process commands.

Make sure that the KA10 patches are after the NIA20 patches since I see changes to kx10_sys.c.

Note I have some patches to kx10_cpu.c that are pending. These effect KL10 paging and RH device allocation routines.

larsbrinkhoff:
I'm rebasing when ever I notice updates to the kl10 branch.

larsbrinkhoff:
Resolves #179.

larsbrinkhoff:
I rebased on top of master, but that destroyed the history.  I'll open a new pull request.

PR #182, opened 2020-02-17
SEL32: Update console device to pass SEL diags.

AZBevier:
SEL32: Update disk devices to hold current STAR in UNIT structure.
SEL32: Add new disk geometry macros.
SEL32: Revise disk formatting support for UTX.
SEL32: Revise INCH command support for all devices.


rcornwell:
Merged in. Diagnostics should exit without any error if it passes.

PR #184, opened 2020-02-18
KA10: Fully working IMX device

larsbrinkhoff:
This adds interrupts to the IMX device, which is needed for ITS to use it.  Also adds SDL game controllers as inputs, for Dazzle Dart to be usable.

This is just a draft pull request.  It need not go in yet.

### DIFF
--- a/PDP10/ka10_imx.c
+++ b/PDP10/ka10_imx.c
@@ -1,6 +1,6 @@
 /* ka10_imx.c: Input multplexor for A/D.
 
-   Copyright (c) 2018, Lars Brinkhoff
+   Copyright (c) 2018,2020, Lars Brinkhoff
 
    Permission is hereby granted, free of charge, to any person obtaining a
    copy of this software and associated documentation files (the "Software"),
@@ -25,6 +25,7 @@
 
 #include <time.h>
 #include "kx10_defs.h"
+#include "sim_video.h"
 
 #ifndef NUM_DEVS_IMX
 #define NUM_DEVS_IMX 0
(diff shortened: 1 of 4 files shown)


rcornwell:
The fprintf's to stderr need to be converted into sim_debug entries. If you need them for debugging, you will probably need them in the future :-)

larsbrinkhoff:
Sure.

larsbrinkhoff:
I was going to test this with Dazzle Dart, but again I had a problem with the 340 display hanging ITS.  I added my pi_vect patch which fixed Spacewar, but to no avail.

larsbrinkhoff:
I needed to add a clr_intrrupt or else ITS would hang when INT and DPY are used at the same time.  However, I still occasionally get the below.

```
KA: APR ERROR IN NULL JOB  
PI LEVEL 7 BUGPAUSE.  FIND A WIZARD, CONSIDER TAKING A CRASH DUMP,
OR TYPE <ALTMODE>P TO ATTEMPT TO REVIVE THE SYSTEM.
YOU ARE NOW IN DDT.
BUGPC/   CNLJL+2   $Q-2/   CONSZ 230220   $P

**Comment by larsbrinkhoff** [comment]
This is now working quite well, except I'd like there to be a way to say how IMX channels are connected to game controllers.

Maybe like this: `set imx channel 73=game0,axis0,negate` to have channel 73 take input from game controller 0, axis 0, and negate the value.

**Comment by larsbrinkhoff** [comment]
Rebased and tested.

**Comment by larsbrinkhoff** [comment]
Blurb:

### A/D Input Multiplexer (IMX) (Device 574)

Provides 128 channels of 12-bit analog to digital converters.  Currently only input from USB game controllers are supported.  To map game inputs to channels, use

```
               SET IMX CHANNEL=123;UNIT0;AXIS1
```

Where `123` is the A/D channel in octal notation, UNIT0 is first USB game controller, and AXIS1 is the second axis on that controller.  Add `;NEGATE` to invert the signal.

PR #185, opened 2020-02-20
KA10: Interface III device with display library.

larsbrinkhoff:
Something like this.  Draft!  Untested!  Actually, I'm quite sure this will not work 100% but it shows how to tie the device to the display library.

If we are to follow the display library pattern, the main logic should be in iii.c rather than ka10_iii.c.

### DIFF
--- a/PDP10/ka10_iii.c
+++ b/PDP10/ka10_iii.c
@@ -32,6 +32,7 @@
 
 #if NUM_DEVS_III > 0 
 #include "display/display.h"
+#include "display/iii.h"
 
 
 #define III_DEVNUM        0430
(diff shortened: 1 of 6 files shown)


larsbrinkhoff:
For keyboard, use this callback.  This will potentially clash with DPY and the "Stanford keyboard".

```
/* A device simulator can optionally set the vid_display_kb_event_process
 * routine pointer to the address of a routine.
 * Simulator code which uses the display library which processes window
 * keyboard data with code in display/sim_ws.c can use this routine to
 * explicitly get access to keyboard events that arrive in the display
 * window.  This routine should return 0 if it has handled the event that
 * was passed, and non zero if it didn't handle it.  If the routine address
 * is not set or a non zero return value occurs, then the keyboard event
 * will be processed by the display library which may then be handled as
 * console character input if the device console code is implemented to
 * accept this.
 */
extern int (*vid_display_kb_event_process)(SIM_KEY_EVENT *kev);

rcornwell:
Thanks this code helped a bunch.

Issue #190, opened 2020-02-24
KA10: Fix 340


larsbrinkhoff:
The 340 has some problems that should be fixed:

- There's a clr_interrupt missing which makes interrupts happen back-to-back.
- With this fixed, the transfer rate is much slower which makes the display flicker.
- The transfer rate can't be too high, or ITS will take notice and temporarily disable the display.
- Consider coscheduling the interrupts to reduce the load on the host.

larsbrinkhoff:
FYI @philbudne

larsbrinkhoff:
ITS has a parameter called MDISWD which is the maximum number of 36-bit words transferred to the 340 per half second.  It's set to 70000 octal, which means the maximum rate is 57344 words per second.  If this limit is reached, ITS will turn off the display for 300 (decimal) clock ticks.  This should be 5 seconds (but seems to be more like one second).

I don't know why this limit is in place.  It corresponds to 8 microseconds per 340 instruction, and almost all isntructions should take much longer.  Maybe it's to prevent really degenerate display lists like all zeros which are parameter mode no-op instructions.

With the current implementations, interrupts aren't cleared so words are sent back-to-back.  This does hit the limit easily which makes the display flicker badly.

larsbrinkhoff:
Fixing the interrupts makes the transfer agree with the DPY device rate which is 500 microseconds per word.  This is a bit too slow for an acceptable refresh rate.  Maybe 100 would be more in the right ballpark, but then the load on the host is quite large.

I'm toying with an idea of a less accurate but less stressful mode where the DPY service calls are "coscheduled" with the clock interrupt running at 60 Hz.  Each such call would bulk transfer the right number of words to display.

larsbrinkhoff:
I have tested my theory.  Fixing interrupt clearing and lowering the interval to 100 microseconds makes the display work much better.  All programs work fine under ITS and standalone, with two exceptions: timesharing Spacewar and SHRDLU TWDEMO.  They both draw and then pause for about one second.  But it's not due to the MDISWD limit.

larsbrinkhoff:
Timing information from PDP-7 340 DISPLAY PROGRAMMING MANUAL.

> The display requires 3 μsec for information transfer, 35 μsec per point for random positioning, and 0.5 μsec per point for intensification.  In vector, vector continue, increment, and character modes the display only requires 1.5 μsec per point for positioning.

philbudne:
I doubt there's another non-DMA interface for the Type 340
[...]

Issue #89, opened 2018-12-15
KA10: 340 vector continuation mode


larsbrinkhoff:
Vector continuation mode behaves strangely near the edge of the display.

![vector](https://user-images.githubusercontent.com/775050/50040452-d9e54180-0043-11e9-9f87-82ce0b242447.png)

CC @philbudne

PR #193, opened 2020-02-27
KA10: Fixes for 340 display

larsbrinkhoff:
Draft.  Work in progress.  See #190, #89.


larsbrinkhoff:
I suggest merging these simple fixes now.  They can possibly be improved on in the future.

larsbrinkhoff:
The first commit is to ensure the device turns the interrupt on or off properly.  Previously there was a problem with interrupts not turning off.

The second commit tunes the interrupt rate to something approximately in the ballpark or the real thing.  This makes the display update rate reasonable.  Note that too fast a rate will make ITS unhappy.

I have tested this with some NTS and ITS programs: Spacewar, MacHack, FED, etc.

PR #195, opened 2020-02-29
KA10: Fix SHOW IMX bug.

larsbrinkhoff:
Tiny bug fix: display channels in octal instead of decimal.  They are input as octal, so decimal is confusing.


rcornwell:
I manually created this commit and merged it since I got another commit on repo. I assigned you as author.

PR #201, opened 2020-03-13
SEL32: Allow multiple disk drive for UTX

AZBevier:


rcornwell:
Manually added this due to remerging previous commit.

PR #221, opened 2020-11-21
KA10: Full screen 340 window

larsbrinkhoff:
I plan to update the 340 window so that F11 toggles full screen.


larsbrinkhoff:
The STK device handles 340 window key events if it's enabled.  If it's not enabled, fall back to a new DPY device key event handler.

PR #229, opened 2021-01-30
KA10: Fix MTY bug.

larsbrinkhoff:
Ensure only one of input and output can interrupt at the same time.


larsbrinkhoff:
This has been rebased.  It should be ready to go in.

PR #252, opened 2021-04-18
KA10: Fix III invivible wholine

larsbrinkhoff:
The III display looks suboptimal.

Actual:
Bad III

Expected:
Good III


larsbrinkhoff:
Experimentally, I determined that the WAITS login screen displays lines up to Y coordinate 522.  This is the new result.
iii-better

PR #260, opened 2021-07-19
KA10: Updates for the AUXCPU and TEN11 devices.

larsbrinkhoff:
First, change order of the TEN11 and AUXCPU address range checks.  AUXCPU should be first so it can shadow part of the TEN11 memory.

Second, add a configurable TEN11 BASE parameter to set the base address of the TEN11 range.


larsbrinkhoff:
Reminder for tonight.

PR #283, opened 2022-04-10
KI10: Include the IMP device.

larsbrinkhoff:
Tiny patch to include the IMP device in the KI10 simulator.


rcornwell:
I did this in a different way. IMP should be enabled for most systems.

PR #285, opened 2022-04-18
KA10: Mask flag request bit with interrupt enable bits.

larsbrinkhoff:
As per table 2-4 on page 2-13, the flag request bit should only be set if the corresponding enable bits are set.

http://www.bitsavers.org/pdf/dec/pdp10/periph/TD10_DECtape_Control_Maint_Jun69.pdf

This makes TENEX read the DECtape directory block correctly.

```
gET FILE dta0:dluser.sav [Confirm]
```

### DIFF
--- a/PDP10/kx10_dt.c
+++ b/PDP10/kx10_dt.c
@@ -504,7 +504,7 @@ t_stat dt_devio(uint32 dev, uint64 *data) {
 
      case CONI|04:
           *data = dtsb;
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
Note, need updated tendmp tool which knows how to write a format acceptable to TENEX:

https://github.com/larsbrinkhoff/pdp10-its-disassembler/pull/143

PR #300, opened 2023-04-22
KA10: Add support for SAIL Data Disc displays

larsbrinkhoff:
Work in progress.  I need to clean things up a bit.

There are known deficiencies.  In particular, scrolling the display leaves some stuff behind.  So far I have not been able to figure out how it's supposed to work.


larsbrinkhoff:
This is ready now.

larsbrinkhoff:
The vds is logically a separate device, but it's only useful for the DD displays.  Maybe it should be invisible to the user, and have `set dd enabled` enable vds automatically?

larsbrinkhoff:
To test: `SET DD ENABLED`

PR #313, opened 2023-11-13
KA10: Faster Chaosnet

larsbrinkhoff:
The CH10 device would poll using clock_coschedule with an interval of 1000.  I'm not sure how many milliseconds this ends up being, but it was rather slow.  I have implemented a new heuristic which polls every 0.3 ms if there are incoming packets.  If not, the old poll interval is used.

Experimenting with e.g. listing a remote tape with DUMP shows this is at least ten times faster than before.

### DIFF
--- a/PDP10/ka10_ch10.c
+++ b/PDP10/ka10_ch10.c
@@ -256,7 +256,7 @@ t_stat ch10_transmit ()
   return SCPE_OK;
 }
 
-void ch10_receive (void)
+int ch10_receive (void)
 {
   size_t count;
   const uint8 *p;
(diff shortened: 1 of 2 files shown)


larsbrinkhoff:
CC @eswenson1 test this branch for speedy KA10 Chaosnet.

rcornwell:
You should make similar changes to ks10_ch11.c

larsbrinkhoff:
Done.  Not tested, but obviously correct.

eswenson1:
I tried this pdp10-ka and it took a very long time to do a full dump of my EXA instance.  Seemed longer than normal, but I could be wrong.  Now, I'm doing a LIST and it is really slow, still.  Perhaps it is somewhat faster than before, but still much slower than KLH10.

larsbrinkhoff:
Thanks for checking.  I will test with pdp10-ks soon.

larsbrinkhoff:
I tested pdp10-ks which was good because I messed it up.  It's fixed now.

PR #316, opened 2024-01-01
PDP6: Add GE DATANET-760 device.

larsbrinkhoff:
This is a terminal multiplexor to emulate the GE DATANET-760 controller.

There are two independent IO bus devices: 70 for input, and 750 for output.  I believe this makes it necessary to make two DEVICE structures.  We only need to expose the GE device to the user, and the internal GTYO device is enabled or disabled automatically.  Maybe there's a better way to accomplish this.

See also https://gunkies.org/wiki/MIT_AI_lab_IO_bus_interface_for_GE_Datanet_760

### DIFF
--- a/PDP10/kx10_defs.h
+++ b/PDP10/kx10_defs.h
@@ -542,6 +542,8 @@ extern DEVICE   dup_dev;
 extern DEVICE   tcu_dev;
 extern DEVICE   ddc_dev;
 extern DEVICE   tym_dev;
+extern DEVICE   ge_dev;
+extern DEVICE   gtyo_dev;
 
 #if KS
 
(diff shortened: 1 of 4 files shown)


larsbrinkhoff:
I updated the GTYO device, hoping to make it invisible:
```
DEVICE gtyo_dev = {
  "GTYO", NULL, NULL, NULL,
  0, 8, 18, 1, 8, 36,
  NULL, NULL, NULL,
  NULL, NULL, NULL,
  &gtyo_dib, DEV_DIS,
  0, NULL,
  NULL, NULL, NULL,
  NULL, NULL, NULL
};
```

The GE reset routine toggles the GTYO enabled/disabled as required.  However, the GTYO device still shows up in the "SHOW DEVICES" list.

PR #318, opened 2024-02-22
CMake: Updates from open-simh PR

bscottm:
simgen/packaging.py customized for this makefile.

(Rich: Full shootin' match in terms of the patchset requested. With the extra video support.)


rcornwell:
Added.

PR #320, opened 2024-04-17
KA10: Fix write access for auxiliary processor memory.

larsbrinkhoff:
They auxiliary memory may shadow a portion of the 10-11 range, so it must be checked first.

### DIFF
--- a/PDP10/kx10_cpu.c
+++ b/PDP10/kx10_cpu.c
@@ -3523,6 +3523,7 @@ int Mem_read_its(int flag, int cur_context, int fetch, int mod) {
                 check_apr_irq();
                 return 1;
             }
+            return 0;
         }
 #endif
 #if NUM_DEVS_TEN11 > 0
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
I previously fixed read access, but forgot about writing! https://github.com/rcornwell/sims/pull/260/commits/bad641393e5868bd5bfb5eb9ffa5f56562e86b6e

larsbrinkhoff:
Steps to repeat:
1. Get https://github.com/aap/pdp6
2. cd pdp6/emu; make
3. Edit init.ini to have netmem refer to localhost, same port (10006).
4. Start ITS with auxcpu enabled on port 10006.  I.e. the default configuration for pdp10-ka off https://github.com/PDP-10/its
5. Start the PDP-6 emulator; just type `./pdp6`
6. Start the PDP-6 machine by toggling power on.
7. Log in to ITS.
8. Type `SYS$J`, then `$$^R`, then `PDP6UP/ -1`
9. Type `PDP6$J`, then `$L .;@ WAR`

At this point, without the fix, the console teletype will say TOO MANY NXM ERRORS.  With the fix, the WAR code will load.  The game can be started by setting address switches to 151 and then pushing the start key.

PR #321, opened 2024-06-16
Replace egrep with grep -E

stevefalco:
Fedora 40 produces warnings that egrep is deprecated.  We can replace it with grep -E, which is functionally identical.

For more information, please see:

https://www.phoronix.com/news/GNU-Grep-3.8-Stop-egrep-fgrep https://github.com/saltstack/salt/issues/65608


larsbrinkhoff:
Ping @rcornwell.

larsbrinkhoff:
Ping again.  Waiting for this to resolve https://github.com/PDP-10/its/issues/2321

rcornwell:
Applied via pull from master.

PR #327, opened 2025-12-08
B5500, I701, I704, I7010, I7070, I7080, I7090: Remove single UNIT DISABLE

markpizz:
If a simh DEVICE has the DEV_DISABLE flag set AND it has 1 UNIT, there really should never be a UNIT_DISABLE flag bit in the UNIT initializer.  The reason for this is that, for many things, the DEVICE name can be used equivalently to represent the single UNIT. What are we trying to achieve here?

- If the DEVICE is disabled, then the unit doesn't exist.  That's fine.

- If the UNIT is disabled (if the UNIT_DISABLE flag is present allowing SET <unit> DISABLE) what would that then mean about the DEVICE?

Bob Supnik's thoughts:

The rationale was that the device represents the controller logic, the unit the physical transducer. So you could disconnect or power down a paper tape reader, but its controller logic would still be there and would still respond in some way.

Mark Pizzolato said:
If the fidelity to the hardware in simulation really behaved precisely as the hardware worked, maybe, but I doubt that this has been a design goal.  Even if it had been a design goal, I see 2 issues:
   a) it would seem that it would have been impossible to produce exactly
      the failure details that real hardware did without a crazy amount
      of side by side testing.
   b) What good would be having a broken device like this connected to a
      simulated system?

### DIFF
--- a/B5500/b5500_urec.c
+++ b/B5500/b5500_urec.c
@@ -31,9 +31,9 @@
 #include "sim_defs.h"
 #include "sim_console.h"
 
-#define UNIT_CDR        UNIT_ATTABLE | UNIT_RO | UNIT_DISABLE | MODE_029
-#define UNIT_CDP        UNIT_ATTABLE | UNIT_SEQ | UNIT_DISABLE | MODE_029
-#define UNIT_LPR        UNIT_ATTABLE | UNIT_SEQ | UNIT_DISABLE
+#define UNIT_CDR        UNIT_ATTABLE | UNIT_RO | ((NUM_DEVS_CDR > 1) ? UNIT_DISABLE : 0) | MODE_029
+#define UNIT_CDP        UNIT_ATTABLE | UNIT_SEQ | ((NUM_DEVS_CDP > 1) ? UNIT_DISABLE : 0) | MODE_029
+#define UNIT_LPR        UNIT_ATTABLE | UNIT_SEQ | ((NUM_DEVS_LPR > 1) ? UNIT_DISABLE : 0)
 
 #define TMR_RTC         0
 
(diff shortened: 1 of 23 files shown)


rcornwell:
I'm going to reject this patch since it does not serve any purpose.

markpizz:
It removes the option to day "SET <unit> DISABLE" for devices which have a single UNIT.

- Please explain how such a device would be useful with it's single unit disabled?
- Please precisely how the simulated system will know this in a way that it did on real hardware?
- 
Certainly, DEVICEs with multiple UNITs could still be useful if some of their UNITs are non-existant, but otherwise, how will you be able to tell precisely how the running simulated system would be able to see exactly what the original hardware saw when the single UNIT is disabled?

    B5500 simulator Open SIMH V4.1-0 Current        git commit id: 4d383732
    sim> help cp set

    CP device SET commands:

    set CP FORMAT                   Sets card format
    set CP ENABLE                   Enables device CP
    set CP DISABLE                  Disables device CP
    set CP DEBUG                    Enables debugging for device CP
    set CP NODEBUG                  Disables debugging for device CP
    set CP DEBUG=CMD;DATA;DETAIL;EXP;CARD
                                    Enables specific debugging for device CP
    set CP NODEBUG=CMD;DATA;DETAIL;EXP;CARD
                                    Disables specific debugging for device CP
    set CP ENABLE                   Enables unit CP
    set CP DISABLE                  Disables unit CP
    set CP APPEND                   Sets CP position to EOF

    *CP device DEBUG settings:
        CMD         Show command execution to devices
        DATA        Show data transfers
        DETAIL      Show details about device
        EXP         Show console data
        CARD        Show Card read/punches

Notice the duplicate SET CP ENABLE SET CP DISABLE options.

rcornwell:
Right now I am knee deep in microsim360. Trying to get reference IBM360 so I can finish my simH IBM360.

I still see this patch as cosmetic.

